Issue #43, opened 2023-04-14
Add components section to improve reusability through referencing within the spec


frankkilcommins:
It's currently possible to compose a workflows document and reference other workflow documents, but referencing capabilities of steps and/or reusing step inputs across workflows or steps is limited.

The current scope and referencing of inputs and outputs for steps and workflows is described by the following picture:
![Workflows-Access-Scope-for-Inputs-and-Outputs](https://user-images.githubusercontent.com/15875424/231983697-766daad7-9345-419c-b915-46f569ea9fa2.png)

Defining workflow inputs, steps and/or step inputs under a global components section (similar to OpenAPI) could bring better reusability through referencing.

##### Components
Holds a set of reusable objects for different aspects of the Workflows Specification. All objects defined within the components object will have no effect on the workflow unless they are explicitly referenced from properties outside the components object.

| Field Name | Type | Description
|:-------------|:------:|-------------
| inputs | Map[`string`, [Schema Object]()] | An object to hold reusable JSON Schema objects to be referenced from workflow inputs. _Remark: possibly trimmed down version of OAS Schema Object_
| ~~steps~~ | ~~Map[`string`, [Step Object](https://mikeralphson.github.io/OpenAPI-Specification/workflows/1.0.0.html#step-object)]~~ | ~~An object to hold a reusable [Step Objects](https://mikeralphson.github.io/OpenAPI-Specification/workflows/1.0.0.html#step-object).~~
| parameters | Map[`string`, [Parameter Object](https://mikeralphson.github.io/OpenAPI-Specification/workflows/1.0.0.html#parameter-object)] | An object to hold reusable [Parameter Objects](https://mikeralphson.github.io/OpenAPI-Specification/workflows/1.0.0.html#parameter-object)

MikeRalphson:
> possibly trimmed down version of OAS Schema Object

@frankkilcommins what would be the rationale for a cut-down schema object? We have got ourselves into a right state in OAS with subset-superset-sidesets before.

shaisachs:
I added reference objects just now. Incorporating them into Steps and Parameters was no problem. However, because the `workflowInputs` is defined to be a `JSON Schema` object, it's a little more difficult - we need to embed reference objects inside that JSON Schema. Can we modify `workflowInputs` to something like an array of JSON Schema or reference object? Something like `[JSON Schema | Reference Object]`.

MikeRalphson:
@shaisachs doesnt the schema object allow `$ref`s anyway?

[...]

PR #51, opened 2023-04-26
Components proposal

shaisachs:
Addresses #43

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -48,11 +48,14 @@ The Workflows Specification can articulate these workflows in a human and machin
       - [Component Object](#component-object)
         - [Fixed Fields](#fixed-fields-6)
         - [Component Object Example](#component-object-example)
-      - [Success Action Object](#success-action-object)
+      - [Reference Object](#reference-object)
         - [Fixed Fields](#fixed-fields-7)
+        - [Reference Object Example](#reference-object-example)
+      - [Success Action Object](#success-action-object)
+        - [Fixed Fields](#fixed-fields-8)
         - [Success Action Object Example](#success-action-object-example)
       - [Failure Action Object](#failure-action-object)
-        - [Fixed Fields](#fixed-fields-7)
+        - [Fixed Fields](#fixed-fields-9)
         - [Failure Action Object Example](#failure-action-object-example)     
     - [Runtime Expressions](#runtime-expressions)
     - [Specification Extensions](#specification-extensions)
@@ -254,7 +257,7 @@ Field Name | Type | Description
 <a name="workflowSummary"></a>summary | `string` | A summary of the purpose or objective of the workflow.
 <a name="workflowDescription"></a>description | `string` | A description of the workflow. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="workflowInputs"></a>inputs | `JSON Schema` or `string` | A JSON Schema object representing the input parameters used by this workflow.  If a `string` is provided, it must correspond to the key of the [[Component Object]](#component-object)'s `inputs` field.
-<a name="workflowSteps"></a>steps | [[Step Object](#step-object)] or `string` | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow.  If a `string` is provided, it must correspond to the key of the [[Component Object]](#component-object)'s `steps` field.
+<a name="workflowSteps"></a>steps | [[Step Object](#step-object) | [Reference Object](#reference-object)] or `string` | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow. If a Reference Object is provided, it MUST refer to a Step.
 <a name="outputs"></a>outputs | Map[`string`, {expression}] |  A map between a friendly name and a dynamic output value. The name MUST use keys that match the regular expression: `^[a-zA-Z0-9\.\-_]+$`.
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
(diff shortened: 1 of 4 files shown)


versions/1.0.0.md
```diff
@@ -257,7 +257,7 @@ Field Name | Type | Description
...
 <a name="workflowSummary"></a>summary | `string` | A summary of the purpose or objective of the workflow.
 <a name="workflowDescription"></a>description | `string` | A description of the workflow. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="workflowInputs"></a>inputs | `JSON Schema` or `string` | A JSON Schema object representing the input parameters used by this workflow.  If a `string` is provided, it must correspond to the key of the [[Component Object]](#component-object)'s `inputs` field.
+<a name="workflowInputs"></a>inputs | `JSON Schema` | A JSON Schema 2020-12 object representing the input parameters used by this workflow. Some of the input parameters may be [References](#reference-object) to the [[Component Object]](#component-object)'s `inputs` field.
```
handrews: If this proposal is trying for the same feel as OAS 3.1, then you don't need to talk about Reference Objects here.  Reference Objects are used  _outside_ of Schema Objects, where `$ref` is not otherwise defined.  The OAS 3.1 Schema Object is just a regular 2020-12 schema, and `$ref` is already defined with behavior specific to schemas (in terms of how adjacent keywords are handled).

shaisachs: Hmm ok, I can take out that sentence. I'll leave the example in though.

versions/1.0.0.md
```diff
@@ -480,6 +487,91 @@ criteria:
...
+
+The `$ref` string value contains a URI [RFC3986](https://tools.ietf.org/html/rfc3986), which identifies the location of the value being referenced.
+
+See the rules for resolving [Relative References](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#relativeReferencesURI).
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="referenceRef"></a>$ref | `string` | **REQUIRED**. The reference identifier. This MUST be in the form of a URI, with a prefix of `#/components/inputs` or `#/components/parameters`.
+<a name="referenceValue"></a>value | `string` | The value of the Input or Parameter which by default SHOULD override that of the referenced component.
```
handrews, line 570: This is a dramatic departure from how `$ref` is used in every other specification I've seen, and will cause problems with anyone trying to use standard tooling.  In particular, if this is done with `$ref` inside of schemas, you will not be able to use standard draft 2020-12 JSON Schema implementations, which was the main reason that OAS 3.1 adopted 2020-12 in full.

I'm kind of dropping in here without context- is there somewhere I can read up on the reason for this usage?   don't see any discussion of it in the referenced issue.

shaisachs: Probably makes sense to discuss at our meeting this coming Wednesday, can you make it? Also, might be worth taking a look at the pet store example which I've marked up in this PR, that should at least convey the intention of this use of $ref. But certainly open to hearing improvements.

handrews: I'll try to make the meeting.  My main concern here is that this proposal seems to be a bit disconnected from the general history and usage of `$ref`.  Support for `$ref` has been a challenge in the OAS tooling community, and has often been addressed by having standalone tools that understand `$ref` which are used in conjunction with OAS-specific tools.  Adding substantially different behavior to `$ref` will break many, many things.

versions/1.0.0.md
```diff
@@ -480,6 +487,91 @@ criteria:
...
+user-name
+my.org.User
+```
+
+##### Component Object Example
+
+```json
+"components": {
+  "inputs": {
+    "storeId": {
```
darrelmiller, line 532: These examples seem to be backwards no?  Isn't input a JSON Schema and parameter has the the name, in value property?

versions/1.0.0.md
```diff
@@ -250,8 +257,8 @@ Field Name | Type | Description
...
 <a name="workflowSummary"></a>summary | `string` | A summary of the purpose or objective of the workflow.
 <a name="workflowDescription"></a>description | `string` | A description of the workflow. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="workflowInputs"></a>inputs | `JSON Schema` | A JSON Schema object representing the input parameters used by this workflow
-<a name="workflowSteps"></a>steps | [[Step Object](#step-object)] | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow
+<a name="workflowInputs"></a>inputs | `JSON Schema` | A JSON Schema 2020-12 object representing the input parameters used by this workflow.
+<a name="workflowSteps"></a>steps | [[Step Object](#step-object) | [Reference Object](#reference-object)] or `string` | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow. If a Reference Object is provided, it MUST refer to a Step.
```
frankkilcommins, line 261: Some formatting issues here. I would state as follows as below if proceeding but I'm not sure if we decided to hold back from referencing _steps_ for now?
```suggestion
<a name="workflowSteps"></a>steps | [[Step Object](#step-object) \| [Reference Object](#reference-object)] | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow. The list can use the [Reference Object](#referenceObject) to link to steps that are defined at the [Workflows Object's components/parameters](#component-object). If a Reference Object is provided, it MUST refer to a Step.

```

shaisachs: Yeah I thought we were knocking out Steps in favor of just using Workflows as the unit of reuse. My pet store example shows how to do that.

versions/1.0.0.md
```diff
@@ -121,6 +127,7 @@ Field Name | Type | Description
 <a name="workflowsInfo"></a>info | [Info Object](#info-object) | **REQUIRED**. Provides metadata about the Workflows. The metadata MAY be used by tooling as required.
 <a name="workflowsSources"></a>sources | [[Source Object](#source-object)] | **REQUIRED**. A list of source documents (such as an OpenAPI document) this workflow SHALL apply to. The list MUST have at least one entry.
 <a name="workflows"></a>workflows | [[Workflow Object](#workflow-object)] | **REQUIRED**. A list of workflows. The list MUST have at least one entry.
+<a name="components"></a>components | [Component Object](#component-object) | An object containing 0 or more components.
```
frankkilcommins: Perhaps keep similar to OAS
```suggestion
<a name="components"></a>components | [Component Object](#component-object) | An element to hold various schemas for the document

```

versions/1.0.0.md
```diff
@@ -480,6 +487,91 @@ criteria:
...
+Field Name | Type | Description
+---|:---|---
+<a name="componentInputs"></a> inputs | Map[`string`, [Schema Object](#schemaObject)] | An object to hold reusable JSON Schema objects to be referenced from workflow inputs.
+<a name="componentParameters">parameters</a> | Map[string, [Parameter Object](#parameter-object)] | An object to hold reusable Parameter Objects
+
+This object MAY NOT be extended with [Specification Extensions](#specificationExtensions).
+
+All the fixed fields declared above are objects that MUST use keys that match the regular expression: `^[a-zA-Z0-9\.\-_]+$`. The key is used to refer to the input or parameter in other parts of the Workflow document.
+
+Field Name Examples:
```
frankkilcommins, line 517: I'm not sure what the field name examples are representing here. We don't have this in other sections

shaisachs: It's just a regex for the keys in the Maps - happy to remove this section but I figured it'd be a good idea.

versions/1.0.0.md
shaisachs: Hmm - components.inputs is a Map of string to JSON Schema object, so I'm reasonably sure that piece is right (in this case `storeId` is the key). Not sure I understand the problem with parameter.

versions/1.0.0.md
```diff
@@ -250,8 +257,8 @@ Field Name | Type | Description
...
 <a name="workflowSummary"></a>summary | `string` | A summary of the purpose or objective of the workflow.
 <a name="workflowDescription"></a>description | `string` | A description of the workflow. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="workflowInputs"></a>inputs | `JSON Schema` | A JSON Schema object representing the input parameters used by this workflow
-<a name="workflowSteps"></a>steps | [[Step Object](#step-object)] | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow
+<a name="workflowInputs"></a>inputs | `JSON Schema` | A JSON Schema 2020-12 object representing the input parameters used by this workflow.
+<a name="workflowSteps"></a>steps | [[Step Object](#step-object) \| [Reference Object](#reference-object)] | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow. If a Reference Object is provided, it MUST refer to a Step.
```
frankkilcommins:
```suggestion
<a name="workflowSteps"></a>steps | [[Step Object](#step-object)] | **REQUIRED**. An ordered list of steps where each step represents a call to an API operation or to another workflow.
```

versions/1.0.0.md
```diff
@@ -480,6 +487,91 @@ criteria:
...
+Holds a set of reusable objects for different aspects of the Workflows Specification. All objects defined within the components object will have no effect on the workflow unless they are explicitly referenced from properties outside the components object.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---|---
+<a name="componentInputs"></a> inputs | Map[`string`, JSON Schema Object] | An object to hold reusable JSON Schema objects to be referenced from workflow inputs.
+<a name="componentParameters">parameters</a> | Map[`string`, [Parameter Object](#parameter-object)] | An object to hold reusable Parameter Objects
+
+This object MAY NOT be extended with [Specification Extensions](#specificationExtensions).
```
frankkilcommins:
```suggestion
This object _MAY_ be extended with [Specification Extensions](#specificationExtensions).

```

frankkilcommins (dismissed):

versions/1.0.0.md
```diff
@@ -480,6 +487,92 @@ criteria:
...
+    }
+  }
+}
+```
+
+#### Reference Object
+
+**Note -** This section is derived from the [OpenAPI Specification, v3.1](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#referenceObject), and is intended to be implemented in similar fashion.
+
+A simple object to allow referencing other components in the OpenAPI document, internally and externally.
```
frankkilcommins:
```suggestion
A simple object to allow referencing other components in the Workflows document.
```

frankkilcommins (approved):

PR #69, opened 2023-09-07
Frankkilcommins/issue66

### LINKED ISSUE #66: Enhance Success and Failure Action Objects to allowing referencing a WorkflowId
Currently, it's only possible to reference a `stepId` as the target for a `goto` type. It would be logical to also allow a `workflowId` to be specified and have it mutually exclusive to a `stepId`.

Logical scenario:
- Having a Failure Action Object instance check if the returned status was `401` and then to reference the _login Workflow Id_ to go off and fetch a fresh auth token

TBD if the context should remain at the current step or have ability to retry current step.

frankkilcommins:
fixes #66


kevin-postman (approved):
All looks good as we had discussed. One thought, might be worth adding a blurb on the retry bit about stepId ALSO being within the same workflow? I think per our discussion about tooling that would also make sense like the Goto.. to keep stepId in scope of workflow? Other than that, looks good.

frankkilcommins:
Thanks @kevin-postman, the last [commit](https://github.com/OAI/sig-workflows/pull/69/commits/8bae8c80c0e177e40103aaf40aa05866196497fb) on the PR adds the restriction on `stepId` 

> The referenced `stepId` SHOULD be within the current workflow

PR #77, opened 2023-10-14
Harden-expression-and-condition-capabilities

### LINKED ISSUE #44: Harden the Runtime Expression section within the spec
Review and harden the [runtime expressions](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#runtime-expressions) documentation.

Expressions are use by:
- input values
- parameter values
- output values

Success and Failure Object criteria also uses expressions and in the draft examples there is usage of JSONPath. Should alternative _expression_ references be used here?

Review usage of JSONPath, JSON Pointer and/or determine if any form of custom expressions (like [GitHub action expressions](https://docs.github.com/en/actions/learn-github-actions/expressions)) would be required to meet the needs of the success/failure criteria expressions. Should this then be split from the [runtime expressions](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#runtime-expressions).

frankkilcommins:
fixes #44


versions/1.0.0.md
```diff
@@ -614,6 +620,74 @@ This object cannot be extended with additional properties and any properties add
...
+`.`| Property de-reference
+
+String comparisons `SHOULD` be case insensitive.
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="criterionContext"></a>context | `{expression}` | A [runtime expression](#runtime-expressions) used to set the context for the condition to be applied on. If `type` is specified, then the `context` MUST be provided (e.g. `$response.body` would set the context that a JSONPath query expression could be applied to).
+<a name="criterionCondition"></a>condition | `string` | **REQUIRED**. The condition to apply. Conditions can be simple (e.g. `$statusCode == 200` which applies a operator on a value obtained from a runtime expression), or a regex, or a JSONPath expression. For regex and [JSONPath](https://datatracker.ietf.org/doc/draft-ietf-jsonpath-base/21/), the `type` and `context` MUST be specified.
+<a name="criterionType"></a>type | `string` | The type of condition to be applied. If specified, the options allowed are `regex` or `JSONPath`. If omitted, then the condition is assumed to be simple, which at most combines literals, operators and [Runtime Expressions](#runtime-expressions).
```
ndenny: I'm in two minds about this myself - but should be add a `simple` type which is the default type, instead of having an implicit simple type if none is specified?

ndenny (approved):
This is great! 
Happy to approve as-is, but left a comment in case anyone else has an opinion?

versions/1.0.0.md
frankkilcommins: I was very much on the fence here. For completeness, it's probably better to have it as part of allowed values. I doubt, I'd ever write it but no harm in allowing folks to be explicit if they want to.

frankkilcommins: extended to now support `simple`

frankkilcommins:
I will update some of the verbiage around Step.successCriteria (to highlight its and AND operation) and also within Action Success/Failure (it's an OR).

frankkilcommins:
@ndenny @kevin-postman I've made the additional changes as we discussed yesterday. If all are happy, let's merge

kevin-postman:
LGTM

ndenny (approved):
👍

PR #93, opened 2023-11-30
Simplify "Reference Objects `$ref` keywords" to "Reference Objects"

ralfhandl:
Increases readability, and the referenced section on reference objects defines the `$ref` keyword.

Also make it plain text as "Reference Objects" is not a keyword.


frankkilcommins (approved):
LGTM

PR #95, opened 2023-11-30
operationRef matching referenced OpenAPI description

ralfhandl:
And the subsequent `parameters` block

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -184,7 +184,7 @@ workflows:
       sessionToken: $response.body
   - stepId: getPetStep
     description: retrieve a pet by status from the GET pets endpoint
-    operationRef: https://petstore3.swagger.io/api/v3/openapi.json#/paths/users/~findbystatus~1{status}/get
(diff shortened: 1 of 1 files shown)


gcatanese:
According to OpenAPI Workflow spec [v1.0.0](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#fixed-fields-4) `"The operation being referenced MUST be described within one of the source descriptions."`.

If I understand correctly this means the `operationRef`  should refer to `./pet-coupons.openapi.yaml` that is mentioned in `sourceDescriptions`

ralfhandl:
> If I understand correctly this means the `operationRef` should refer to `./pet-coupons.openapi.yaml` that is mentioned in `sourceDescriptions`

Good point, so not only the fragment part of the example operationRef needs repair, the path part also has room for improvement 😄

ralfhandl:
@frankkilcommins Please review

frankkilcommins (approved):
LGTM. Thanks for these fixes @ralfhandl and yes as @gcatanese the operations being referenced should be located within a `sourceDescription`

PR #111, opened 2023-12-20
chore: force case insensitivity for string comparisons

### LINKED ISSUE #108: Update Criterion Object to enforce case insensitivity
Section [Criterion Object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#criterion-object) says:

> String comparisons SHOULD be case insensitive.

Please make this a MUST for either sensitive or insensitive, don't care much which.

Different behavior across different workflow execution engines would be a nuisance.

_Originally posted by @ralfhandl in https://github.com/OAI/sig-workflows/issues/49#issuecomment-1835897435_

frankkilcommins:
fixes #108

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -657,7 +657,7 @@ Operator | Description
 `[]`| Index (0-based)
 `.`| Property de-reference
 
-String comparisons `SHOULD` be case insensitive.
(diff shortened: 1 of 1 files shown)


kevin-postman (approved):
LGTM

PR #125, opened 2024-01-26
clarify how to reference `operationId` and `workflowId` in case of clashes

### LINKED ISSUE #124: Clarify how step `operationId` and/or `workflowId` should be prefixed if multiple `sourceDescriptions` are defined
Section [Step Object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#step-object), field `operationId`
> If more than one (non `workflowsSpec` type) source description is defined within a Workflows Description, then the `operationId` specified MUST be prefixed with the source name to avoid ambiguity or potential clashes.

How prefixed? Just prepend without separator character? An example would be useful.

Same with field `workflowId`.

_Originally posted by @ralfhandl in https://github.com/OAI/sig-workflows/issues/49#issuecomment-1834131310_

frankkilcommins:
fixes #124


handrews (commented):
This looks good - I might suggest creating a section up front that explains how various references and other connections are resolved.  That would pull together the information that is duplicated in several spots here with other things we spoke about the other day such as referencing into full documents.

ndenny (approved):

frankkilcommins:
> This looks good - I might suggest creating a section up front that explains how various references and other connections are resolved. That would pull together the information that is duplicated in several spots here with other things we spoke about the other day such as referencing into full documents.

I'll review deduplicating some of the verbiage as part of a separate issue

PR #142, opened 2024-02-07
Swap inputs and parameters to match schema

### LINKED ISSUE #140: Components Object Example incorrect
It appears that Components Object [Example](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#components-object-example) is incorrect. 

Components `inputs` is defined as map of Parameter objects while `parameters` is defined as map of Schema, shouldn't that be the other way around?

```
components:
  inputs:
    storeId:
      name: storeId
      in: header
      value: $inputs.x-store-id
  parameters:
    pagination:
      type: object
      properties:
        page:
          type: integer
          format: int32
        pageSize:
          type: integer
          format: int32
```

ndenny:
Fixes #140

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -548,12 +548,12 @@ my.org.User
 
 ```yaml
 components:
-  inputs:
+  parameters:
     storeId:
       name: storeId
       in: header
       value: $inputs.x-store-id
-  parameters:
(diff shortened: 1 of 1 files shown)


frankkilcommins (approved):
nice spot

Issue #139, opened 2024-02-03
Parameter target JSON Pointer


gcatanese:
The [Parameter Object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#parameter-object) is supposed to be a valid JSON Pointer. 

In the Pet Coupons [example](https://github.com/OAI/sig-workflows/blob/e021569488ee7109e4dc1b272f9d204b932e238e/examples/1.0.0/pet-coupons.workflow.yaml#L119) the `target` is defined as `$request.body#/petId`, which is not a valid JSON Pointer.

In the FAPI-PAR [example](https://github.com/OAI/sig-workflows/blob/e021569488ee7109e4dc1b272f9d204b932e238e/examples/1.0.0/FAPI-PAR.workflow.yaml#L153) the `target` is defined as `/grant_type`, which is a valid JSON Pointer.

I guess the notation `$request.body#/json/pointer` makes clear where the target applies and it is flexible (other target options eventually?). Is it then better to clarify the usage of `target` in the Parameter Object? Or decide to apply a formal JSON Pointer implementation, therefore updating the examples?

handrews:
The current wording for `target` includes:

> Useful for targeting specific request body parts

It sounds like that ought to be strengthened to say that JSON Pointers MUST be resolved against the request body.  Or it should allow full runtime expressions if there are other valid locations.

gcatanese:
Indeed. It would be great to avoid any ambiguity. 

**Option 1**
The `target` attribute is a valid JSON Pointer (i.e. `/status`) resolved to the request body. In this case, it MUST be specified when `in` is defined as`body`.

**Options 2**
The `target` attribute is an expression (i.e. `$request.body#/status`) resolving to the request body, or other valid location.

Option 2 gives more flexibility, while option 1 gives clear guidance (that suits tooling better).

ndenny:
Yeah, I think we have to go for Option 1. Option 2 may have just been an old idea we missed when updating the examples. I'll prep a quick PR to make it match for now.

ndenny:
Gave it a go in #143

frankkilcommins:
The intention here is indeed to use JSON Pointers to be resolved against the request body. As of now, I don't see a use case for targeting other locations. Thus we can improve the verbiage a little.

[...]

PR #143, opened 2024-02-07
Fix target values to be valid JSON Pointers

ndenny:
Assuming we go for option 1 in #139 and follow the spec

### DIFF
--- a/examples/1.0.0/pet-coupons.workflow.yaml
+++ b/examples/1.0.0/pet-coupons.workflow.yaml
@@ -116,22 +116,22 @@ workflows:
         parameters:
           - name: pet_id
             in: body
-            target: $request.body#/petId
+            target: /petId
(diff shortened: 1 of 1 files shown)


gcatanese:
Should the specification of [Parameter object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#parameter-object) also be modified to indicate that `target` MUST be specified when `in` is defined as `body`?

frankkilcommins:
> Should the specification of [Parameter object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#parameter-object) also be modified to indicate that `target` MUST be specified when `in` is defined as `body`?

@gcatanese see comment https://github.com/OAI/sig-workflows/issues/139#issuecomment-1935648977

Issue #130, opened 2024-01-26
Review `operationRef` templating verbiage


frankkilcommins:
From @handrews feedback [review](https://github.com/OAI/sig-workflows/issues/49#issuecomment-1836661780):

Regarding `operationRef`:
> A relative or absolute URI reference to an OAS operation. This field is mutually exclusive of the operationId and workflowId fields respectively. A complete [URI Template](https://www.rfc-editor.org/rfc/rfc6570) SHOULD be used. The operation being referenced MUST be described within one of the source descriptions.

Why is this a URI Template? Why is it different from OAS templating? I philosophically approve of using standard URI Templates, but you're using them with a Parameter Object including style which is designed for OAS templating (the Parameter Object explicitly references OAS templating, and makes no mention of RFC 6570 at all). And it is unclear how other features of URI Templates are intended to interact with the Parameter Object. I realize this was probably discussed somewhere in this repository, but it's surprising enough that I think it needs addressing in the specification. I would not understand how to implement this as it is.

gcatanese:
> From the meeting - suggest we enforce use of $sourceDescriptions - so: `operationRef: $sourceDescriptions.<name>#/path/to/operation` [or is it `operationRef: $sourceDescriptions.<name>.url#/path/to/operation` ]
> 
> which would expand for tooling to be a URI as `($sourceDescriptions.<name>.url) + #path/to/operation`

Would it make sense to enforce it for `operationId` too (i.e. `$sourceDescriptions.<name>.<operationId>`)?

handrews:
> `operationRef: $sourceDescriptions.<name>#/path/to/operation`
> [or is it `operationRef: $sourceDescriptions.<name>.url#/path/to/operation` ]

OK so just to be clear, this `operationRef` is intentionally going to behave differently from the OAS Link Object's `operationRef`?

It looks to me like the objective is to avoid using URIs directly and instead always use imported `sourceDescription` identifiers.  I think this is a _very good idea_.  Moonwalk is also moving towards named imports.

Might I suggest something like `operationPath` instead?  This captures the fact that you need a JSON Pointer path (presumably because the operation does not have an `operationId`), and avoids the cognitive dissonance of `operationRef` meaning something slightly different here than in OAS.

Since you are using plain text JSON Pointers elsewhere, it would make sense to use them for `operationPath` as well, instead of using the URI fragment syntax with its extra layer of escaping.

ndenny:
Gave it shot in #144

[...]

PR #144, opened 2024-02-07
Specification Path expressions for operations with no OperationI ID

ndenny:
As discussed in #130  - a slight modification on @handrews suggestion - I'm using the `#` to separate the `$sourceDescriptions.<name>` from the JSON Pointer. This disambiguates it from the `operationId` syntax that uses `$sourceDescriptions.<name>.<operationId>`

Note sure if I needed to create a completely separate ABNF - let me know.


handrews:
@ndenny I see what you mean about the syntactic ambiguity, but if you're going to use `#` then you should use the URI fragment-encoded syntax of JSON Pointers.  Because if it looks like a URI fragment, it should parse like a URI fragment.

I'm also not sure that the `#` really disambiguates it.  The OAS does not define syntax for `operationId`, it is just RECOMMENDED to follow "common programming language naming conventions".  That doesn't forbid `/`, but it doesn't forbid `#` either.  Unless people have been starting `operationId`s with `/` then it should be just as distinctive.  Although you do have to remember that the empty string is a valid JSON pointer, and technically also not forbidden as an `operationId`...

I see you have the `#` followed by plain-text JSON Pointer elsewhere as well.  I'm concerned that's going to cause very obscure bugs when someone assumes they should URI-encode the pointer, as the result is also a valid JSON Pointer... just not the _same_ JSON Pointer if you interpret it as plain text.  That error can be hard to spot and debug.

If you can't do something with the runtime expression syntax to disambiguate it, I think it either needs to be URI-encoded or you need a different delimiter.

ndenny:
@handrews - perhaps it'd be cleaner if we had a pair of properties:
`operationSourceDescription` and `operationPointer` so we don't have to 'encode' what we mean into one thing

handrews:
@ndenny yeah... same applies to `requestBody`, I think.  Either way, if we're not uri-encoding we shouldn't use `#` as it just looks like a URI-fragment and folks will expect it to use that encoding.

frankkilcommins:
`sourceDescriptions` is already covered by the ABNF syntax specified for [runtime expressions](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#runtime-expressions). No need to create another section.

I don't have a strong opinion on the property naming. The obvious choices are `operationRef` (understanding it deviates from OAS), `operationReference`, or `operationPath` (as suggested by @handrews )

How I envisaged the URI reference to be created was to leverage the runtime expression to specify the base and then to use JSON Pointer to resolve against the appropriate path operation.

Given:
```
sourceDescriptions:
- name: petStoreDescription
  url: https://github.com/swagger-api/swagger-petstore/blob/master/src/main/resources/openapi.yaml
  type: openapi
```

The we could reference an operation in the file via:
`'{$sourceDescriptions.petStoreDescription.url}#/paths/~1pet~1findByStatus/get'`

frankkilcommins (changes requested):
See comment https://github.com/OAI/sig-workflows/pull/144#issuecomment-1936216343

handrews:
> How I envisaged the URI reference to be created was to leverage the runtime expression to specify the base and then to use JSON Pointer to resolve against the appropriate path operation.

My concern about the `#` usage is that given this JSON:

```JSON
{
  "cat": {
    "Schrödinger": "maybe"
  }
}
```
We should either see `"/cat/Schrödinger"` (plain string) or `"#/cat/Schr%C3%B6dinger"` (urlencoded).  We should not mix the URI-fragment separator `#` with the plain string JSON Pointer text.  Anything that looks like a fragment should parse like a fragment.

But I'd argue that in runtime expressions, it's better to use a non-fragment syntax.  Particularly for things like requestBody that might get applied to a plain `application/json` payload.  _Technically_ you can only evaluate a fragment based on the rules of the media type, and `application/json` specifically _does not_ define JSON Pointer as a fragment syntax (and the JSON Pointer RFC specifically mentions that it is not the fragment syntax for `application/json`).  So it's actually more correct to define runtime expression-specific rules using plain text JSON Pointers rather than URI fragments.

handrews:
.... and now I'm realizing that `#` + JSON Pointer (string form) is already in the runtime expression syntax of OAS 3.x :-/  So I guess we can file that under "things Workflows is not expected to fix."  We can fix it with Moonwalk and I guess Workflows 2.  Workflows 1 should match OAS 3 where possible.

handrews (approved):

frankkilcommins:
We've decided on `operationPath`. @ndenny to remove the redundant ABNF section that was added

ndenny:
Question - if we're going for the expected syntax to be `'{$sourceDescriptions.petStoreDescription.url}#/paths/~1pet~1findByStatus/get'` ? Do we need to then explain the curly-brackets syntax?

frankkilcommins (approved):

PR #145, opened 2024-02-07
Remove references to WHATWG to avoid confusion

### LINKED ISSUE #127: Revisit URL verbiage in Source Descriptions
From @handrews review [feedback](https://github.com/OAI/sig-workflows/issues/49#issuecomment-1836661780).

The spec defines:
> _A URL to a source description to be used by a Workflow. MUST be in the form of a [URL](https://url.spec.whatwg.org/). MAY be defined using relative references as defined by [RFC3986](https://spec.openapis.org/oas/latest.html#bib-RFC3986)._

Referencing WHATWG here is problematically inconsistent with OAS 3.x. If you must reference WHATWG (a decision with which I rather disagree, because that spec is insanely hard to interpret unless you're writing a parser exactly the way they want you to, although I understand perhaps I'm too late with this feedback), "URL" is not the correct term as it is defined in terms of in-memory structs. The correct term would be "URL String". I'm still doubtful that WHATWG's "URL String" and RFC 3986 are entirely compatible, and that strikes me as a problem given that RFC 3986 is used everywhere else in the OpenAPI world. And you're even mixing WHATWG's "URL" definition with RFC 3986's relative reference resolution process, which is not necessarily identical to WHATWG's relative resolution process (I haven't really checked that part, though).

Since WHATWG's URL spec is a "Living Specification", this requirement may effectively change without notice, further introducing interoperability problems. "Living specifications" are a radically different model than how OpenAPI specs work.

If the desire here is purely to ensure that the URL is resolvable in accordance with its scheme, you can just say that along with referencing RFC 3986. I strongly believe that OpenAPI as a whole, across all of its specifications, should consistently use either IETF RFCs 3986/3987/6874 or WHATWG, and not mix the two. Preferably the IETF RFCs as that is what is already established.

With most URI/URL-related libraries you're lucky if you can find one that [properly conforms to any specification](https://github.com/handrews/iritest). Finding one that will conform to WHATWG for some things and RFC 3986 for others is almost certainly impossible. Except perhaps by coincidence.

Also, you probably want an "and" between "URL," and "MAY be defined...".

ndenny:
Fixes #127

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -237,14 +237,15 @@ version: 1.0.1
 
 Describes a source description (such as an OpenAPI description) that will be referenced by one or more workflows described within a Workflows Description.
 
-An object storing a map between named description keys and location URLs to the source descriptions (such as an OpenAPI description) this workflow SHALL apply to. Each source location `string` MUST be in the form of a URL and MAY be defined using relative references as defined by [RFC3986](https://spec.openapis.org/oas/latest.html#bib-RFC3986).
+An object storing a map between named description keys and location URLs to the source descriptions (such as an OpenAPI description) this workflow SHALL apply to. Each source location `string` MUST be in the form of a URL. 
+If a relative reference is used it MUST be in the form of a URI-reference as defined by [RFC3986 section 4.1](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2).
 
 ##### Fixed Fields
 
 Field Name | Type | Description
 ---|:---:|---
 <a name="sourceName"></a>name | `string` | **REQUIRED**. A unique name for the source description. Tools and libraries MAY use the `name` to uniquely identify a source description, therefore, it is RECOMMENDED to follow common programming naming conventions. SHOULD conform to the regular expression `[A-Za-z0-9_\-]+`.
-<a name="sourceURL"></a>url | `string` | **REQUIRED**. A URL to a source description to be used by a Workflow. The URL MAY be defined using relative references as defined by [RFC3986](https://spec.openapis.org/oas/latest.html#bib-RFC3986).
+<a name="sourceURL"></a>url | `string` | **REQUIRED**. A URL to a source description to be used by a Workflow. If a relative reference is used, it MUST be in the form of a URI-reference as defined by [RFC3986 section 4.1](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2).
 <a name="infoType"></a>type | `string` | The type of source description. Possible values are `"openapi"` or `"workflowsSpec"`.
 
 
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -237,14 +237,14 @@ version: 1.0.1
 
 Describes a source description (such as an OpenAPI description) that will be referenced by one or more workflows described within a Workflows Description.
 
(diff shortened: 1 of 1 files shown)


versions/1.0.0.md
```diff
@@ -238,14 +238,14 @@ version: 1.0.1
 
 Describes a source description (such as an OpenAPI description) that will be referenced by one or more workflows described within a Workflows Description.
 
-An object storing a map between named description keys and location URLs to the source descriptions (such as an OpenAPI description) this workflow SHALL apply to. Each source location `string` MUST be in the form of a [URL](https://url.spec.whatwg.org/). MAY be defined using relative references as defined by [RFC3986](https://spec.openapis.org/oas/latest.html#bib-RFC3986).
+An object storing a map between named description keys and location URLs to the source descriptions (such as an OpenAPI description) this workflow SHALL apply to. Each source location `string` MUST be in the form of a URL and MAY be defined using relative references as defined by [RFC3986](https://spec.openapis.org/oas/latest.html#bib-RFC3986).
```
handrews: Thanks for settling on RFC 3986!  I'd be fine with this, but you could change it to:

> MUST be in the form of a URI-reference as defined by RFC3986 section 4.1.

You have the "location URL" phrase which is the correct term for the usage, and "URI-reference" is the correct term for the syntax.

frankkilcommins:
I think if the comment from @handrews can be applied then this looks good. 

IMO we should reference RFC3986 Section 4.2 directly for relative references (https://datatracker.ietf.org/doc/html/rfc3986#section-4.2) rather than via the OAS latest reference table.

versions/1.0.0.md
frankkilcommins: IMO we should reference RFC3986 Section 4.2 directly for relative references (https://datatracker.ietf.org/doc/html/rfc3986#section-4.2) rather than via the OAS latest reference table.

frankkilcommins: There's two places where this can be resolved

versions/1.0.0.md
```diff
@@ -238,14 +238,15 @@ version: 1.0.1
 
 Describes a source description (such as an OpenAPI description) that will be referenced by one or more workflows described within a Workflows Description.
 
-An object storing a map between named description keys and location URLs to the source descriptions (such as an OpenAPI description) this workflow SHALL apply to. Each source location `string` MUST be in the form of a URL and MAY be defined using relative references as defined by [RFC3986](https://spec.openapis.org/oas/latest.html#bib-RFC3986).
+An object storing a map between named description keys and location URLs to the source descriptions (such as an OpenAPI description) this workflow SHALL apply to. Each source location `string` MUST be in the form of a URL. 
+If a relative reference is used it MUST be in the form of a URI-reference as defined by [RFC3986 section 4.1](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2).
```
handrews: With apologies, I'm going to have to be pedantic here:  "URI-reference" (§4.1) includes both "URI" (with a scheme; §1.1.1) and "relative reference" (§4.2).  Also, you mention 4.1 in the text and 4.2 in the link.

Really, you can just say that the value is a URI-reference (§4.1) and that covers everything.

frankkilcommins: @ndenny could you make the little update suggested?

ndenny:
@frankkilcommins - finally fixed :)

frankkilcommins (approved):

PR #147, opened 2024-02-09
Clarify Parameter target JSON Pointer

### LINKED ISSUE #139: Parameter target JSON Pointer

frankkilcommins:
Fixes #139


ndenny (approved):
Yeah, that makes sense. We can't have a JSON Pointer into a path, query, header or cookie.

handrews (approved):

PR #148, opened 2024-02-09
add `in` property to parameters

frankkilcommins:


ndenny (approved):

gcatanese:
@frankkilcommins @ndenny Is this change correct? The attribute `in` has been added, but `type` and `format` are not Parameter attributes.

```
  parameters:
    page:
      in: query
      type: integer
      format: int32
```

frankkilcommins:
@gcatanese true they should not be specified. I missed that when cleaning up. There's new issue out there to fix that

PR #151, opened 2024-02-15
Added README to provide details on existing examples and provide addi…

kevinduffey:
Adding this to provide details on the examples so consumers of the spec can learn how to use the spec from the examples.


ndenny (approved):
I'm pretty sure I originated the OAuth one and I vaguely know what Mark Haine was trying to do with the FAPI one, so I'll try remember what they show! 😅

PR #158, opened 2024-03-01
Chore: fix `pet-coupons.workflow.yaml` example

### LINKED ISSUE #154: Potential bug in the `pet-coupons.workflow.yaml` example
This looks like a bug in the example:

https://github.com/OAI/sig-workflows/blob/9164da1e3e63df94737d84bac4b135e2479688a8/examples/1.0.0/pet-coupons.workflow.yaml#L162-L170

I believe it can't have `type` and `format ` fields and instead it should have `name`, `value` and`in` fields.

frankkilcommins:
Fixes #154


kevinduffey (approved):
LGTM

PR #159, opened 2024-03-01
feat: remove support for `dependsOn` for v1.0.0

### LINKED ISSUE #153: `dependsOn` clarification and parallel execution
We have a question about `dependsOn`?

Why is it required? In the description it says it's to **ensure steps are "not processed in parallel"** BUT `steps` are defined as **An ordered list of steps**.

There is no mentions of parallel vs sequential execution.

Do we really need parallel execution complexity in the first version of the spec?
I'm worried that `dependsOn` may add too much friction for the workflow author.

frankkilcommins:
fixes #153


kevinduffey (approved):
LGTM

PR #162, opened 2024-03-13
feat: Add Request Body Object

### LINKED ISSUE #152: Request body design update?
We find the current design of request body not very flexible and not consistent with OpenAPI 3.0 definition terminology. (we hit issues trying to cover our real-life use cases)

Are there any reasons for this specific design?

We have an alternative design which we want to share here and get some feedback!

The main idea is to describe request body params in requestBody property (same as OpenAPI 3.0) instead of in: body (was used in OpenAPI 2.0).

We can work on the formal definition but would be faster to start from a few examples to gather feedback first.

Example from the spec:

```yaml
- stepId: do-the-auth-flow
  requestBody:
    # contentType is optional, by default should use the one defined in OpenAPI spec or application/json
    # Object value will be serialized according to the contentType if possible
    contentType: application/x-www-form-urlencoded
    value:
      client_id: $inputs.clientId
      grant_type: $inputs.grantType
      redirect_uri: $inputs.redirectUri
      client_secret: $inputs.clientSecret
      code: $steps.browser-authorize.outputs.code
      scope: $inputs.scope
```

We can control the serialization format with [OpenAPI encoding object](https://swagger.io/specification/#encoding-object), e.g.:

```yaml
- stepId: do-the-auth-flow
  requestBody:
    contentType: application/x-www-form-urlencoded
    encoding:
      scope:
        style: form
        explode: true
    value:
      client_id: $inputs.clientId
      grant_type: $inputs.grantType
      redirect_uri: $inputs.redirectUri
      client_secret: $inputs.clientSecret
      code: $steps.browser-authorize.outputs.code
      scope: $inputs.scope
```

Application/json example:

```yaml
- stepId: place-order
  requestBody:
    contentType: application/json
    value:
      pet_id: $steps.find-pet.outputs.my_pet_id
      coupon_code: $steps.find-coupons.outputs.my_coupon_code
```

Example with XML body:

```
- stepId: place-order-xml
  operationId: placeOrderXml
  requestBody:
    contentType: application/xml
    value: |
      <xml>
        <petId>{$inputs.pet_id}</petId>
        <couponCode>{$inputs.coupon_code}</couponCode>
        <quantity>{$inputs.quantity}</quantity>
        <status>placed</status>
        <complete>false</complete>
      </xml>
```

We also have an idea for file uploads using special `$file`:

```yaml
- stepId: upload-receipt-with-status
  operationId: uploadReceipt
  requestBody:
    contentType: multipart/form-data
    encoding:
      file:
        contentType: application/pdf
    value: 
      file: $file('./receipt.pdf') # uses special $file to load file from the file system, if encoding is not defined, it will be encoded as application/octet-stream
      status: placed
```

What do you think about this approach? We like it more as it is more flexible (covers all the real use-cases we and our users have) and also because it reuses terminology from OpenAPI.

Waiting for your feedback!

frankkilcommins:
fixes #152 

Hoping to omit encoding for now and delegate that to the referenced OpenAPI descriptions

### DIFF
--- a/examples/1.0.0/oauth.workflow.yaml
+++ b/examples/1.0.0/oauth.workflow.yaml
@@ -53,11 +53,10 @@ workflows:
           This is where you do the refresh
         operationId: get-token
         requestBody:
+          contentType: application/x-www-form-urlencoded
           payload:
-            contentType: application/x-www-form-urlencoded
-            payload:
-              grant_type: refresh_token
-              refresh_token: $steps.do-the-auth-flow.outputs.my_refresh_token
+            grant_type: refresh_token
+            refresh_token: $steps.do-the-auth-flow.outputs.my_refresh_token
         successCriteria:
           - condition: $statusCode == 200
           - context: $response.body
--- a/examples/1.0.0/oauth.workflow.yaml
+++ b/examples/1.0.0/oauth.workflow.yaml
@@ -52,9 +52,12 @@ workflows:
         description: >-
           This is where you do the refresh
         operationId: get-token
-        parameters:
-          grant_type: refresh_token
-          refresh_token: $steps.do-the-auth-flow.outputs.my_refresh_token
+        requestBody:
+          payload:
+            contentType: application/x-www-form-urlencoded
+            payload:
+              grant_type: refresh_token
+              refresh_token: $steps.do-the-auth-flow.outputs.my_refresh_token
         successCriteria:
           - condition: $statusCode == 200
           - context: $response.body
@@ -89,20 +92,12 @@ workflows:
         description: >-
           This is where you get the token
         operationId: get-token
-        parameters:
-        - name: client_id
-          in: body
-          style: form
-          value: $inputs.client_id
-        - name: client_secret
-          in: body
-          style: form
-          value: $inputs.client_secret
-        - name: grant_type
-          in: body
-          style: form
-          value: 'client_credentials'
-
+        requestBody:
+          contentType: application/x-www-form-urlencoded
+          payload:
+            client_id: $inputs.client_id
+            client_secret: $inputs.client_secret
+            grant_type: client_credentials
         successCriteria:
           - condition: $statusCode == 200
           - context: $response.body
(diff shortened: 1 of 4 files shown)


examples/1.0.0/oauth.workflow.yaml
```diff
@@ -52,9 +52,12 @@ workflows:
         description: >-
           This is where you do the refresh
         operationId: get-token
-        parameters:
-          grant_type: refresh_token
-          refresh_token: $steps.do-the-auth-flow.outputs.my_refresh_token
+        requestBody:
+          payload:
+            contentType: application/x-www-form-urlencoded
+            payload:
```
kevinduffey: Do we want payload: again here? Not sure if Body might be better? Not bothered either way just noticing payload twice.

examples/1.0.0/oauth.workflow.yaml
```diff
@@ -52,9 +52,12 @@ workflows:
         description: >-
           This is where you do the refresh
         operationId: get-token
-        parameters:
-          grant_type: refresh_token
-          refresh_token: $steps.do-the-auth-flow.outputs.my_refresh_token
+        requestBody:
+          payload:
```
kevinduffey, line 57: Actually (posting this AFTER my comment right below).. below you show a content-type: and payload: so I am thinking the payload above content-type is wrong?

examples/1.0.0/pet-coupons.workflow.yaml
```diff
@@ -113,26 +110,15 @@ workflows:
...
-          - name: status
-            in: body
-            target: /status
-            value: "placed"
-          - name: complete
-            in: body
-            target: /complete
-            value: false
+        requestBody:
+          payload:
```
kevinduffey, line 115: Again.. should this be here? isnt it content-type: and payload: (in the same tab location) as opposed to two payload: ??

kevinduffey (commented):
Left a couple of comments.. waiting on those to approve

examples/1.0.0/oauth.workflow.yaml
frankkilcommins: whoops - copy and past error

kevinduffey (approved):
LGTM

PR #182, opened 2024-04-25
Consolidate reusable parameters and reusable objects.

### LINKED ISSUE #181: 3 ways to reference reusable components
After reviewing the last draft of the spec I noticed there are 3 different ways to reference reusable components:

- [Reusable parameter object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#reusable-parameter-object)
- [Reusable objects](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#reusable-object)
- Standard JSON Schema `$ref` for inputs

> A simple object to allow referencing of objects contained within the [Components Object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#components-object) from locations within steps or workflows in the Workflows Description. This excludes parameters and inputs defined in the [components/parameters](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#components-object). Note - parameters should be referenced using the [Reusable Parameter Object](https://github.com/OAI/sig-workflows/blob/main/versions/1.0.0.md#reusable-parameter-object) and inputs should use standard JSON Schema referencing using the $ref keyword.

We believe that the end users (authors) will be constantly confused about which to use in which context.

What problem do we solve with those 3 ways of referencing and why we can't use `$ref`s for all of those (which users are familiar with as well as tooling support).

I believe it has to do with $ref semantics in context of JSON schema.
I believe while it may improve the experience for tooling authors it makes it much worse for the end users.

But I'm not even sure if it makes it easier for the tooling authors as now we have to deal with more error messages and validations.

-----

On a separate note I'm not clear why the Reusable objects has this specific structure:

```
{
    "name": "{$components.successActions.notify}"
}
```

The moment that I'm unclear about is why the field is called `name` while the expression does not resolve to the name or anything like name?

frankkilcommins:
Simplified the referencing capabilities by consolidating `Reusable Parameter Object` into `Reusable Object`. This now means that there are two referencing mechanisms supported by the Workflows Specification:
- JSON Schema references using `$ref` keyword. This can only be used for JSON Schema - e.g. Inputs defined in components
- non JSON Schema referencing using the `Resuable Object` reference expression - e.g. all other component objects

closes #181


versions/1.0.0.md
```diff
@@ -281,10 +278,10 @@ Field Name | Type | Description
...
+<a name="workflowParameters"></a>parameters | [[Parameter Object](#parameter-object) \| [Reusable Object](#reusable-object)] | A list of parameters that are applicable for all steps described under this workflow. These parameters can be overridden at the step level but cannot be removed there. Each parameter MUST be passed to an operation or workflow as referenced by `operationId`, `operationPath`, or `workflowId` as specified within each step. If a Reusable Object is provided, it MUST link to parameters defined in the [components/parameters](#components-object) of the current Workflows document. The list MUST NOT include duplicate parameters.
```
kevinduffey: "MUST link to parameters defined in." and below (line 338) you have "MUST link to a parameter". I would assume those should be the same wording? I assume a Reusable Object is single and should have one reference to one parameter (as stated in line 338)?

versions/1.0.0.md
```diff
@@ -613,52 +610,39 @@ components:
 }
 ```
 
-#### Reusable Parameter Object
+#### Reusable Object
 
-A simple object to allow reusable parameters from multiple steps or workflows in the Workflows Description.
+A simple object to allow referencing of objects contained within the [Components Object](#components-object). It can be leveraged from locations within steps or workflows in the Workflows Description. **Note** - Input Objects MUST use standard JSON Schema referencing via the `$ref` keyword while all non JSON Schema objects will leverage this object and it's expression based referencing mechanism.
```
adamaltman:
```suggestion
A simple object to allow referencing of objects contained within the [Components Object](#components-object). It can be used from locations within steps or workflows in the Workflows Description. **Note** - Input Objects MUST use standard JSON Schema referencing via the `$ref` keyword while all non JSON Schema objects use this object and its expression based referencing mechanism.

```
Isn't this going to confuse people that they can use `$ref` to reference most components in OpenAPI 3.1, 3.0, 2.0? Yet here, they have to distinguish if it's an input object? Is there a family of objects named input objects? (If so, it didn't stand out to me.)

versions/1.0.0.md
```diff
@@ -613,52 +610,39 @@ components:
...
 
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="reusableParameterName"></a>name | `{expression}` | **REQUIRED**. A [runtime expression](#runtime-expressions) used to reference the desired parameter.
-<a name="reusableParameterValue"></a>value | `string` | A value by default MUST override that of the referenced parameter. 
+<a name="reusableObjectName"></a>reference | `{expression}` | **REQUIRED**. A [runtime expression](#runtime-expressions) used to reference the desired object.
+<a name="reusableObjectValue"></a>value | `string` | A value which by default MUST override that of the referenced parameter. This is only applicable for parameter object references.
```
adamaltman: What does it mean "by default"? Are those words necessary? I wonder why this only applies to parameter object references?
```suggestion
<a name="reusableObjectValue"></a>value | `string` | Sets a value of the referenced parameter. This is only applicable for parameter object references.

```

kevinduffey (approved):

versions/1.0.0.md
```diff
@@ -613,52 +610,39 @@ components:
...
-  name: $components.successActions.notify
+  reference: $components.parameters.page
+  value: 1
 ```
 
 ```json
   {
-    "name": "{$components.successActions.notify}"
+    "reference": "{$components.parameters.page}",
+    "value": 1
```
ndenny: Suggest we make these consistent - JSON and YAML examples - this implies to me that JSON must have a `{` which is not true. (Same with the example above)

ndenny (changes requested):

ndenny (approved):

PR #183, opened 2024-04-25
Remove mutual exclusivity constraint between `payload` and `pointers`. Rename to `replacements`

### LINKED ISSUE #180: Clarifications on `pointers`
Hey @frankkilcommins !

This is a follow-up discussion to https://github.com/OAI/sig-workflows/issues/152 and https://github.com/OAI/sig-workflows/pull/162.

I've noticed you added `pointers` which is mutually exclusive to the `payload` as per the latest version of the spec.

I have concerns about it. 

First, I think JSON pointers can't be used for constructing the JSON Object because of the ambiguity with arrays. For example:

```
pointers:
 -  target: /0
    value: 22
```

Should it result in:

```json
[22]
```

or 

```json
{
  "0": 22
}
```

I also find this line contradicting: "which MUST be resolved against the request body."
Is it a mistake and it should not be mutually exclusive? Could you clarify the intention here.

Also, I find the `pointers` name to be misleading. When I first saw it I couldn't understand it. What about some other names like `patches`, `updates`?

But in general, what problem are we solving with this? What can be accomplished using it which is not possible or hard with the `payload`?

frankkilcommins:
closes #180

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -69,9 +69,9 @@ The Workflows Specification can articulate these workflows in a human-readable a
       - [Request Body Object](#request-body-object)
         - [Fixed Fields](#fixed-fields-12)
         - [Request Body Object Example](#requestbody-object-example)
-      - [Payload Pointer Object](#payload-pointer-object)
+      - [Payload Replacement Object](#payload-replacement-object)
         - [Fixed Fields](#fixed-fields-13)
-        - [Payload Pointer Object Example](#payload-pointer-object-example)        
+        - [Payload Replacement Object Example](#payload-replacement-object-example)        
     - [Runtime Expressions](#runtime-expressions)
     - [Specification Extensions](#specification-extensions)
     - [Security Considerations](#security-considerations)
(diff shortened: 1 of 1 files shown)


versions/1.0.0.md
```diff
@@ -769,8 +769,8 @@ A single request body describing the `Content-Type` and request body content to
...
-<a name="requestBodyPayload"></a>payload | Any | A value representing the request body payload. The value can be a literal value or can contain [Runtime Expressions](#runtime-expressions) which MUST be evaluated prior to calling the referenced operation. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. This field is mutually exclusive to `pointers`.
-<a name="requestBodyPointers"></a>pointers | [[Payload Pointer Object](#payload-pointer-object)] | A list of locations and values to set within a payload. This field is mutually exclusive to `payload`.
+<a name="requestBodyPayload"></a>payload | Any | A value representing the request body payload. The value can be a literal value or can contain [Runtime Expressions](#runtime-expressions) which MUST be evaluated prior to calling the referenced operation. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary.
```
kevinduffey: "that cannot naturally represented in JSON or YAML" probably should say "that cannon naturally be represented in JSON or YAML"

adamaltman:
```suggestion
<a name="requestBodyPayload"></a>payload | Any | A value representing the request body payload. The value can be a literal value or can contain [Runtime Expressions](#runtime-expressions) which MUST be evaluated prior to calling the referenced operation. To represent examples of media types that cannot be naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary.

```

kevinduffey (approved):

ndenny (approved):

PR #198, opened 2024-06-05
Draft PR for Arazzo schema

LasneF:
this is a draft PR, it contains a basic schema that validate the samples
it is mainly reproducing the structure of Arazzo 

it is not complete and should not merge , it misses the reference mechanism that i will pick up from OAS schema

and also few points are set to object , but may be we can do better

### DIFF
--- a/schemas/v1.0/schema.json
+++ b/schemas/v1.0/schema.json
@@ -1,0 +1,347 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Arazzo Specification",
+    "type": "object",
+    "properties": {
+      "arazzo": {
+        "type": "string",
+        "description": "The version number of the Arazzo Specification that the Arazzo Description uses."
+      },
+      "info": {
+        "type": "object",
+        "properties": {
+          "title": {
+            "type": "string",
+            "description": "A human-readable title of the Arazzo Description."
+          },
+          "summary": {
+            "type": "string",
+            "description": "A short summary of the Arazzo Description."
+          },
+          "description": {
+            "type": "string",
+            "description": "A description of the purpose of the workflows defined."
+          },
+          "version": {
+            "type": "string",
+            "description": "The version identifier of the Arazzo document."
+          }
+        },
+        "required": ["title", "version"]
+      },
+      "sourceDescriptions": {
+        "type": "array",
+        "items": {
+          "$ref": "#/$defs/SourceDescription"
+        },
+        "minItems": 1
+      },
+      "workflows": {
+        "type": "array",
+        "items": {
+          "$ref": "#/$defs/Workflow"
+        },
+        "minItems": 1
+      },
+      "components": {
+        "type": "object",
+        "properties": {
+          "inputs": {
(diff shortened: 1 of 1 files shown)


handrews:
@LasneF it would be better to use draft 2020-12 as that's the current draft being used in current versions of OAS and Arazzo.  It's also a lot easier to re-use schemas (see the 3.0 vs 3.1 OAS schemas to see the difference, although I have no idea if it will make as much of a difference with Arazzo).  With the OAS schemas, they use the same draft as is referenced in the spec they describe.

handrews:
@LasneF also, thanks for starting this work!  I should have said that before diving in with feedback 😅

LasneF:
I ve done a basis , but i would need a review , and especially on the points  below 

How to modelize 
* inputs that is describe as a  Map[string, JSON Schema]

```javascript
          "inputs": {
            "type": "object",
            "additionalProperties": {
              "type" : "object"
            },
            "description": "An object to hold reusable JSON Schema objects to be referenced from workflow inputs."
          },
``` 
* I do not have modelize the runtime expression with a regex , 

* when the spec mentionned JSON schema it is not clear if it is inline or if it can be a $ref (that would be nice) 

* Re usable object ,
I ma not super confortable with the definition here 
https://spec.openapis.org/arazzo/latest.html#reusable-object 

vs the OAI definition 
https://spec.openapis.org/oas/latest.html#reference-object 
the usage of the field reference toward $ref 

for now skipping but it will be an easy addition then , i will follow the pattern  if else 
https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json#L350

Sintayew4:
#198

LasneF:
@jetro4u  you are right, dropped , as it more simple . i am not sure how to express that it has be conform to json schema , description may be good enough

handrews:
@LasneF @jetro4u the `additionalProperties: {type: object}` would have been correct as far as it went if it were `additionalProperties: {type: [object, boolean]}`, as schemas can be objects or booleans and not any other type.

The correct way to handle this is shown in the OAS 3.1 schemas and relies on `$dynamicRef` and `$dynamicAnchor`.

LasneF:
@handrews i can reproduce the same model that is on the OAS 3.1 model 
https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml#L726 howerver i do not know if 
1. it would be better to point to OAS schema (looks wrong to me as creating a strong dependency) 
2. use the same concept in arazzo spec . What about the $comment  ? that is pointing to OAS spec 
2.1 would it be fair to keep it as a weak dependency to OAS project )
2.2 just drop it as there is no such definition in Arrazzo spec
3. just put type: ```additionalProperties: {type: [object, boolean]}``` , to keep it simple as for now no reuse 

by the way thanks for pointing the OAS spec i realize that OAS did a great job by adding a $comment on most of the item will add it

LasneF:
closing this PR follow up is on https://github.com/OAI/Arazzo-Specification/pull/224

PR #215, opened 2024-07-18
Minor example updates from linter

lornajane:
I've been using the examples as we worked on adding Arazzo linting to Redocly CLI and I found a couple of things that I _think_ it's correct to fix. Thanks for sharing these examples, they really helped us get started!


kevinduffey (approved):
LGTM

frankkilcommins (approved):

Issue #223, opened 2024-07-30
Schema dialects undefined in Arazzo


jeremyfiel:
Do schema dialects need some mention in the Arazzo specification? 

taken from OAS, I believe this should also apply to Arazzo.

> 4.8.24.2.3 Specifying Schema Dialects
It is important for tooling to be able to determine which dialect or meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema dialect, or some custom meta-schema.

> The $schema keyword MAY be present in any root Schema Object, and if present MUST be used to determine which dialect should be used when processing the schema. This allows use of Schema Objects which comply with other drafts of JSON Schema than the default Draft 2020-12 support. Tooling MUST support the [OAS dialect schema id](https://spec.openapis.org/oas/v3.1.0#dialectSchemaId), and MAY support additional values of $schema.

> To allow use of a different default $schema value for all Schema Objects contained within an OAS document, a jsonSchemaDialect value may be set within the [OpenAPI Object](https://spec.openapis.org/oas/v3.1.0#openapi-object). If this default is not set, then the OAS dialect schema id MUST be used for these Schema Objects. The value of $schema within a Schema Object always overrides any default.

> When a Schema Object is referenced from an external resource which is not an OAS document (e.g. a bare JSON Schema resource), then the value of the $schema keyword for schemas within that resource MUST follow [JSON Schema rules](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8.1.1).

frankkilcommins:
Being honest, this was somewhat _parked_ prior to 1.0.0 delivery. @handrews can you remember anything else from our conversations on it and if defining an explicit dialect for Arazzo is beneficial? 

There are indeed many cases where an Arazzo document can describe workflows over APIs described by various versions of OAS. Instructions for tooling on how to parse/validate OAD's is driven by the OAD and the version of OAS therein, rather than Arazzo having an opinion on that.

Issue #222, opened 2024-07-30
Schema object is not defined in the specification


jeremyfiel:
I believe Arazzo should also have a section for `#schema-object` which defines the same as the OpenAPI. 

> The OpenAPI Schema Object [dialect](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.3.3) is defined as requiring the [OAS base vocabulary](https://spec.openapis.org/oas/v3.1.0#fixed-fields-20), in addition to the vocabularies as specified in the JSON Schema draft 2020-12 [general purpose meta-schema](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8).

The OpenAPI Schema Object dialect for this version of the specification is identified by the URI https://spec.openapis.org/oas/3.1/dialect/base (the “OAS dialect schema id”).

The following properties are taken from the JSON Schema specification but their definitions have been extended by the OAS:

description - [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
format - See [Data Type Formats](https://spec.openapis.org/oas/v3.1.0#dataTypeFormat) for further details. While relying on JSON Schema’s defined formats, the OAS offers a few additional predefined formats.
In addition to the JSON Schema properties comprising the OAS dialect, the Schema Object supports keywords from any other vocabularies, or entirely arbitrary properties.

---

 do not copy this text verbatim because it references the OAS dialect that is not used in Arazzo.  The `description` and `format` keywords seem to follow the same behavior as OAS

frankkilcommins:
We chose not to have a Schema Object so as to steer clear of sub-set/super-set situations.

Issue #221, opened 2024-07-30
runtime expression validation


jeremyfiel:
Is there a common runtime expression validation format or regex out in the wild?  I tried the googles, but nothing comes up.

handrews:
@jeremyfiel you could use [abnf-to-regexp](https://pypi.org/project/abnf-to-regexp/) to generate one.

adamaltman:
ChatGPT proposes this:

```
^\$(?:url|method|statusCode|(?:request|response|message)\.(?:header\.[!#$%&'*+\-.^_`|~0-1]+|query\.[A-Za-z0-9_.-]+|path\.[A-Za-z0-9_.-]+|body(?:#(?:[^\/~]|~[01])*) )|inputs\.[A-Za-z0-9_.-]+|outputs\.[A-Za-z0-9_.-]+|steps\.[A-Za-z0-9_.-]+|workflows\.[A-Za-z0-9_.-]+|sourceDescriptions\.[A-Za-z0-9_.-]+|components\.[A-Za-z0-9_.-]+|components\.parameters\.[A-Za-z0-9_.-]+)$
```

Not confident in that. Also, not confident what CHAR should mean (here it is limited to alphanumeric and `_.-` characters. 

See #253

char0n:
Hi @jeremyfiel.

There is now https://www.npmjs.com/package/@swaggerexpert/arazzo-runtime-expression, Arazzo Runtime Expressions parser, validator and extractor. Hope you and others find it useful.

PR #224, opened 2024-07-30
JSON Schema 2020-12 definition for Arazzo descriptions

### LINKED ISSUE #176: JSON Schema for the specification
We should publish a JSON Schema to help tooling validate specifications, etc...

jeremyfiel:
closes #176

There are three pending issues related to finalizing this draft
#223 
#222 
#221


frankkilcommins:
@jeremyfiel thanks for the initiative here. There is already a draft PR #198 from @LasneF for #176. Can I encourage you both to work together and consolidate under a single PR for review?

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,525 @@
...
+        description: The named location of the parameter
+        enum:
+          - path
+          - query
+          - header
+          - cookie
+          - body
+      value:
+        description: The value to pass in the parameter
+        type: string
```
LasneF: not sure that this must be a string , this matter especially if we handle json , is it a must to have " 

for instance look over there 
https://github.com/OAI/Arazzo-Specification/blob/main/examples/1.0.0/pet-coupons.arazzo.yaml#L74

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,525 @@
...
+  workflow-object:
+    $comment: https://spec.openapis.org/arazzo/v1.0.0#workflow-object
+    description: Describes the steps to be taken across one or more APIs to achieve an objective
+    type: object
+    properties:
+      workflowId:
+        description: Unique string to represent the workflow
+        $dynamicAnchor: workflowId
+        type: string
+        pattern: '^[A-Za-z0-9_\-]+$'
```
LasneF: given the statement 
" it is RECOMMENDED to follow common programming naming conventions. SHOULD conform to the regular expression [A-Za-z0-9_\-]+." 
i am not sure if  the json schema should enforce good practices or not 

the issue over there is that spec that does not follow this , are valid as it is just a SHOULD , whereas leveraging this schema with strict validation it might be rejected ... not sure as pattern is not a mandatory to follow as well

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,525 @@
...
+                properties:
+                  in: true
+    $ref: '#/$defs/specification-extensions'
+    unevaluatedProperties: false
+  request-body-object:
+    $comment: https://spec.openapis.org/arazzo/v1.0.0#request-body-object
+    description: The request body to pass to an operation as referenced by operationId or operationPath
+    type: object
+    properties:
+      description:
```
LasneF: https://github.com/OAI/Arazzo-Specification/blob/main/versions/1.0.0.md#request-body-object there is no description mentionned there

schemas/v1.0/schema.yaml
jeremyfiel: that's fair for a Parameter object.  I updated it to all types. 

The example you provided is invalid because a `reusableObject>value` can only be a string. 

https://spec.openapis.org/arazzo/v1.0.0#reusable-object

schemas/v1.0/schema.yaml
jeremyfiel: this is consistent with the OpenAPI Specification.  They enforce this pattern on component and schemas names and use the same language

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,525 @@
...
+        description: A list of workflows that MUST be completed before this workflow can be processed
+        type: array
+        uniqueItems: true
+        items:
+          type: string
+      steps:
+        description: An ordered list of steps where each step represents a call to an API operation or to another workflow
+        type: array
+        uniqueItems: true
+        minItems: 1
```
LasneF, line 114: not sure if it is required to have min item = 1 
we could this is excluding people putting 
sucessActions : []  that looks valid also ,  notice this pattern is on several other array

schemas/v1.0/schema.yaml
jeremyfiel: although, you can argue that the text indicates the `value` is only for `Parameter Object` references, which indeed takes ANY value.  So we need the spec maintainers to review that behavior

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,525 @@
...
+    description: |-
+      Describes a single workflow step which MAY be a call to an
+      API operation (OpenAPI Operation Object or another Workflow Object)
+    type: object
+    properties:
+      stepId:
+        description: Unique string to represent the step
+        $dynamicAnchor: stepId
+        type: string
+        pattern: '^[A-Za-z0-9_\-]+$'
```
LasneF, line 166: it is a recommendation so should it be in the schema 

honeslty we would set a MUST it would be nice as well

LasneF (commented):
Feedback that apply globaly : 

- should we endforce Id syntax knowing that it is just a recommendation 
- shoud we enforce array to have at least one element vs the syntax [] that is good as well as explicit 
-

schemas/v1.0/schema.yaml
jeremyfiel: `steps` are required in the `workflows` array.  I don't think it's reasonable to assume `steps` can be empty

LasneF: sorry i miss click the location  the subject was for successAction and failureActions

i think having sucessActions : []  or failureActions : [] can be seen as a practice , some people prefer to be very explicit by putting empty array

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,562 @@
+$id: 'https://spec.openapis.org/arazzo/1.0/schema/2024-08-01'
+$schema: 'https://json-schema.org/draft/2020-12/schema'
+description: |-
+  The description of OpenAPI Initiative Arazzo v1.0.0 documents
+  without schema validation, as defined by https://spec.openapis.org/arazzo/v1.0.0
+type: object
+properties:
+  arazzo:
```
DmitryAnansky, line 8: Maybe it makes sense to use enums.
Smth like:
```
{ type: 'string', enum: ['1.0.0'] },
```

jeremyfiel: Th reason I used a pattern is it allows newer versions to be created without a breaking change, particularly for code generation  

This is also consistent with OpenAPI

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+    unevaluatedProperties: false
+  reusable-object:
+    $comment: https://spec.openapis.org/arazzo/v1.0.0#reusable-object
+    description: A simple object to allow referencing of objects contained within the Components Object
+    type: object
+    properties:
+      reference:
+        description: A runtime expression used to reference the desired object
+        type: string
+      value:
```
DmitryAnansky, line 448: In spec this value is `string`

```
value | string | Sets a value of the referenced parameter. This is only applicable for parameter object references.
-- | -- | --
```

https://spec.openapis.org/arazzo/latest.html#fixed-fields-8

frankkilcommins: @DmitryAnansky FYI - There's an open issue for this https://github.com/OAI/Arazzo-Specification/issues/231 as it should really match what is specified at [4.6.6.1Fixed Fields](https://spec.openapis.org/arazzo/latest.html#fixed-fields-4) for `value`

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+$schema: 'https://json-schema.org/draft/2020-12/schema'
+description: |-
+  The description of OpenAPI Initiative Arazzo v1.0.0 documents
+  without schema validation, as defined by https://spec.openapis.org/arazzo/v1.0.0
+type: object
+properties:
+  arazzo:
+    description: The version number of the Arazzo Specification
+    type: string
+    pattern: '^1\.\d+\.\d+(-.+)?$'
```
frankkilcommins:
```suggestion
    pattern: '^1\\.0\\.\\d+(-.+)?$'
```
suggest limiting to 1.0.x

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+        type: string
+        format: uri-reference
+      type:
+        description: The type of source description
+        enum:
+          - arazzo
+          - openapi
+    required:
+      - name
+      - type
```
frankkilcommins:
```suggestion
      - url
```
It's `name` and `url` that are the required properties.

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+        type: array
+        uniqueItems: true
+        minItems: 1
+        items:
+          $ref: '#/$defs/step-object'
+      successActions:
+        description: A list of success actions that are applicable for all steps described under this workflow
+        type: array
+        uniqueItems: true
+        minItems: 1
```
frankkilcommins:
```suggestion
```
optional - no requirement to specify an item

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+        items:
+          oneOf:
+            - $ref: '#/$defs/success-action-object'
+            - $ref: '#/$defs/reusable-object'
+      failureActions:
+        description: A list of failure actions that are applicable for all steps described under this workflow
+        type: array
+        uniqueItems: true
+        minItems: 1
```
frankkilcommins:
```suggestion
```
optional - no requirement to specify an item

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+        description: A map between a friendly name and a dynamic output value
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            type: string
+      parameters:
+        description: A list of parameters that are applicable for all steps described under this workflow
+        type: array
+        uniqueItems: true
+        minItems: 1
```
frankkilcommins:
```suggestion
```
optional - no requirement to specify an item

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+    properties:
+      contentType:
+        description: The Content-Type for the request content
+        type: string
+      payload: true
+      replacements:
+        description: A list of locations and values to set within a payload
+        type: array
+        uniqueItems: true
+        minItems: 1
```
frankkilcommins:
```suggestion
```
optional - no requirement to specify an item

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+    allOf:
+      - if:
+          required:
+            - operationId
+          properties:
+            operationId: true
+        then:
+          properties:
+            parameters:
+              items:
```
frankkilcommins: This should be a oneOf as the item could be a Parameter Object or a Reusable Object.

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,563 @@
...
+                $ref: '#/$defs/parameter-object'
+      - if:
+          required:
+            - operationPath
+          properties:
+            operationPath: true
+        then:
+          properties:
+            parameters:
+              items:
```
frankkilcommins: This should be a oneOf as the item could be a Parameter Object or a Reusable Object.

frankkilcommins (changes requested):

schemas/v1.0/schema.yaml
jeremyfiel: Regarding this `allOf` definition related to `workflowId`, `operationId` and `operationPath`, the specification indicates there are conditions to be met.  That's the reason I created these conditional statements.

https://github.com/OAI/Arazzo-Specification/blob/df64e14d36e4c963256cbdf1b4f3759d8ccd95c5/versions/1.0.0.md?plain=1#L391

> The name location of the parameter. Possible values are "path", "query", "header", "cookie", or "body". When the step in context specifies a workflowId, then all parameters map to workflow inputs. In all other scenarios (e.g., a step specifies an operationId), the in field MUST be specified.

---

My interpretation of this section means if `operationId` or `operationPath` are the type defined, the reference can **only** be a Parameter object and the `in` attribute must be required.

If a `workflowId` is defined, either `parameterObject` or `reusableObject` may be defined.  

Is that a correction interpretation?

schemas/v1.0/schema.yaml
notEthan: I don't see anything in the OpenAPI schema/spec that is like this.
The closest I see, `operationId`, does not specify a pattern.
Other patterns correspond to MUSTs, e.g. "All the fixed fields declared above are objects that MUST use keys that match the regular expression: `^[a-zA-Z0-9\.\-_]+$`." I may be overlooking something but I don't see any that don't.
I think that the schema should not reject values that are not forbidden, even if recommended against; they are still valid.

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,558 @@
...
+    $ref: '#/$defs/specification-extensions'
+    unevaluatedProperties: false
+  workflow-object:
+    $comment: https://spec.openapis.org/arazzo/v1.0.0#workflow-object
+    description: Describes the steps to be taken across one or more APIs to achieve an objective
+    type: object
+    properties:
+      workflowId:
+        description: Unique string to represent the workflow
+        $dynamicAnchor: workflowId
```
notEthan: I'm not seeing how either `workflowId` or `stepId` below would have dynamic behavior - it seems like this should just be an `$anchor` used with `$ref`. But it is entirely possible I am missing something - am I?

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,558 @@
...
+      - required:
+          - operationPath
+      - required:
+          - workflowId
+    allOf:
+      - if:
+          required:
+            - operationId
+          properties:
+            operationId: true
```
notEthan: I don't think `properties` is needed / does anything here?

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,558 @@
...
+        type: array
+        uniqueItems: true
+        minItems: 1
+        items:
+          $ref: '#/$defs/criterion-object'
+    allOf:
+      - if:
+          properties:
+            type:
+              const: goto
```
notEthan, line 369: Should this `if` have `required: [goto]` ?

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,558 @@
...
+        uniqueItems: true
+        items:
+          $ref: '#/$defs/criterion-object'
+    allOf:
+      - if:
+          properties:
+            type:
+              enum:
+                - goto
+                - retry
```
notEthan, line 421: Should this `if` have `required: [type]` ?

schemas/v1.0/schema.yaml
jeremyfiel: that's fair.  it doesn't necessarily need to be dynamic, especially within the same schema. the scope doesn't change.

schemas/v1.0/schema.yaml
jeremyfiel: you're right. I think I had it there while building the conditional and didn't remove it

schemas/v1.0/schema.yaml
jeremyfiel: should be `type`.  i'll add it

EDIT: 
now I know why I didn't have it in the `if`, it's already required at the root of that schema. 
[#L448](https://github.com/OAI/Arazzo-Specification/pull/224/files#diff-c3b79a68b0d389644801b367021ceb2169c0a1afbdec651107b11605db4671c2R448)

schemas/v1.0/schema.yaml
jeremyfiel: I didn't exactly follow their pattern, which now that I look at it closer, I probably could update it to match 100% to reduce some duplication in this one.

but they enumerate all the `components` names and then use `propertyNames` to enforce it. 
https://github.com/OAI/OpenAPI-Specification/blob/ba055ca00cbd735dceef4ed7c9db024cd9bfcd1a/schemas/v3.1/schema.yaml#L199

jeremyfiel: EDIT: 
opps. completely misread the original comment from Francois and then interpreted your comment in the same context as my original interpretation. I see what you're referring to now

i suppose `stepId` and `workflowId` can be relaxed to `type: string` since a `MUST` is not defined

i will defer to the spec authors to comment here. @frankkilcommins

schemas/v1.0/schema.yaml
notEthan: er, yeah, I meant type. I see, that makes sense.

schemas/v1.0/schema.yaml
frankkilcommins: see https://github.com/OAI/Arazzo-Specification/issues/236 (sorry missed this ping)

schemas/v1.0/schema.yaml
```diff
@@ -0,0 +1,545 @@
...
+$schema: 'https://json-schema.org/draft/2020-12/schema'
+description: |-
+  The description of OpenAPI Initiative Arazzo v1.0.0 documents
+  without schema validation, as defined by https://spec.openapis.org/arazzo/v1.0.0
+type: object
+properties:
+  arazzo:
+    description: The version number of the Arazzo Specification
+    type: string
+    pattern: '^1\.0\.\d+(-.+)?$'
```
jeremyfiel, line 11: @frankkilcommins   should this allow `-dev`, `-beta` etc... on the arrazo version?  I was looking at your implementation for Spectral and you have it limited to [0-9]*. 

https://github.com/stoplightio/spectral/pull/2663/files#diff-3df514595919dcfbae1a8420f9bcbd88f7987073e79b965c1e1367bb959f10cfR6

--- 

do we want to align on the proper regex here?  Yours is slightly different than mine too.  Are you using non-capture group for performance reasons or something else?

schemas/v1.0/schema.yaml
frankkilcommins: It's best to treat this as a `SHOULD` indicates a strong recommendation but it is **not** mandatory. Higher level tools/linters can provide mechanisms to _warn_ or _enforce_ the strong recommendation based user preferences.

schemas/v1.0/schema.yaml
frankkilcommins: > i think having sucessActions : [] or failureActions : [] can be seen as a practice , some people prefer to be very explicit by putting empty array

Correct - there is no minItems requirement. I believe this has already been addressed

schemas/v1.0/schema.yaml
frankkilcommins: @jeremyfiel allowing `-dev`, `-rc1`, etc. can make sense from an Arrazo project perspective as we're working on enhancements to the specification.

Spectral is more targeted at authors downstream of a published specification version. We would not expect to publish something outside of what's covered by `1\.0(?:\.[0-9]*)?$`. That being said, it was more of a Spectral choice to validate in that way and I followed that convention above.

frankkilcommins (approved):
LGTM - Thanks @jeremyfiel

frankkilcommins:
@jeremyfiel could you add a `schema.json` as part of this too?

jeremyfiel:
In the OAS repo, I believe they use yaml as the base schema and there is a JSON version generated during the build. Is that something Arazzo repo will replicate?  If not, I can a JSON version too.

frankkilcommins:
I think it's manually maintained - see example: https://github.com/OAI/OpenAPI-Specification/commits/main/schemas/v3.0/schema.json

There's a job that generates JSON examples for YAML examples, but I don't see similar for schemas.

handrews:
@frankkilcommins we definitely only take PRs on the YAML schemas.  The schema publishing process is completely broken, though.  See OAI/OpenAPI-Specification#3715.  It's been broken for years but there's never been anyone with both the time and ability to fix it.  In theory we'll fix it after the patch releases ship but idk who's going to do the work on it.

LasneF (approved):

frankkilcommins:
@handrews in the interest of simplicity we'll mandate changes on both the JSON and YAML schemas for now.....and time permitting we'll just require one and generate the other.

frankkilcommins (approved):

ndenny (approved):
Thanks for the work!

PR #278, opened 2024-11-14
Fix reference to the CoC

earth2marsh:
Because the OAI has an org-level code of conduct, this repo picks that up by default. IIUC, is preferable to refer to this using the project's instantiation of that document.


frankkilcommins (approved):
Thanks @earth2marsh

PR #289, opened 2024-11-22
respec build script: create minor-version symlinks

ralfhandl:
The minor-version symlink points to the latest published patch version and allows linking to the minor version, for example from the patch-version-independent JSON schema for Arazzo descriptions.


frankkilcommins (approved):
Nice!

PR #290, opened 2024-11-25
schema-publish workflow

ralfhandl:
- replace WORK-IN-PROGRESS placeholder in YAML schema with last change date
- rename schema to that date
- create JSON counterpart
- create pull request for `gh-pages` branch of OAS repo

### DIFF
--- a/.github/workflows/schema-publish.yaml
+++ b/.github/workflows/schema-publish.yaml
@@ -1,0 +1,59 @@
+name: schema-publish
+
+# author: @ralfhandl
+
+#
+# This workflow copies the x.y schemas to the gh-pages branch
+#
+
+# run this on push to main
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch: {}
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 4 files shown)


Iho82 (approved):

.github/workflows/schema-publish.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+#
+# This workflow copies the x.y schemas to the gh-pages branch
+#
+
+# run this on push to main
+on:
+  push:
+    branches:
+      - main
```
frankkilcommins, line 13: Only needs to run if changes to schema directory

```
  push:
    paths:
      - 'schemas/**'
    branches:
      - main
```

ralfhandl: I did not restrict the file pattern to also have it run on changes to the workflow files and its dependencies, and was too lazy to list them all.

Which is not really a problem because the created pull request will only reflect changes to the files-to-be-published, and if there are no changes, nothing happens.

frankkilcommins (approved):
LGTM

PR #291, opened 2024-11-25
feat: add GH action  markdown linting for spec file

jeremyfiel:


package.json
```diff
@@ -37,6 +38,7 @@
     },
     "devDependencies": {
         "mdv": "^1.3.4",
-        "yaml": "2.4.2"
+        "yaml": "2.4.2",
+        "markdownlint-cli": "^0.27.0"
```
jeremyfiel: i also tried updating the version to match `0.42.0` which is what I had originally, no change.  also the script uses npx, so it should grab the latest anyways. This entry is more for local development.

jeremyfiel:
> I split the markdown lint but now the behavior is totally different.. can't figure out why. I added all the same config files and shell scripts. when I ran it previously, it auto-fixed the table pipes but now it's not making those changes and throwing the errors
> 
> i'm baffled.

fixed it.

.github/workflows/validate-markdown.yaml
```diff
@@ -1,10 +1,10 @@
 name: validate-markdown
 
-# Author: @MikeRalphson
-# Issue: https://github.com/OAI/OpenAPI-Specification/issues/2130
+# Author: @ralfhandl based on work of @MikeRalphson
+# Issue: https://github.com/OAI/Overlay-Specification/issues/53
```
ralfhandl: Either remove or refer to an Arazzo issue

jeremyfiel:
updated to run against the `spec` and `schemas readme` files

.github/workflows/validate-markdown.yaml
frankkilcommins: Do we want a dedicated Arazzo issue for this? I feel we either should refer to an Arazzo one or just remove the issue reference.

frankkilcommins:
@jeremyfiel I'm assuming we expect that the validate-markdown will pass here?

jeremyfiel:
ok, this should be ready!  that was more difficult than it needed to be. lol

frankkilcommins:
@jeremyfiel `1.0.1` changes have been merged into **v1.0-dev**. Can you merge in latest / rebase and we'll get this PR included?

jeremyfiel:
not sure why this glob isn't working correctly.  it should ignore `v1.0.0` and not throw the linting errors. 

any help from someone else is appreciated 

```yaml
- name: Lint markdown
      run: npx --yes markdownlint-cli --config .markdownlint.yaml versions/1.0.[^0].md versions/1.[1-9].*.md versions/2.*.md
```

PR #296, opened 2024-12-17
spec build workflow: dynamically determine latest spec version

ralfhandl:
- Use release tag to determine latest spec version


frankkilcommins (approved):
LGTM - thanks @ralfhandl

PR #302, opened 2024-12-22
Workflows with app-generated access token

ralfhandl:
Instead of using personal access tokens the two workflows for publishing HTML specs and JSON Schemas now use an app-generated temporary access token.

Pull requests generated by the modified workflows:
* https://github.com/OAI/OpenAPI-Specification/pull/4275
* https://github.com/OAI/OpenAPI-Specification/pull/4274

### DIFF
--- a/.github/workflows/respec.yaml
+++ b/.github/workflows/respec.yaml
@@ -19,7 +19,7 @@ on:
 jobs:
   respec:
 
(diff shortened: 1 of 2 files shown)


frankkilcommins (approved):
Thanks @ralfhandl

PR #325, opened 2025-02-17
Clarify origin or runtime expressions ABNF non-terminals

### LINKED ISSUE #253: Runtime expressions ABNF clarifications
Should `name` be at least one character?

Example: ABNF defines  a `"$steps." name` as a possible runtime expression. Are expressions like `$steps.` valid? Or does there have to be at least one character after the "$steps."?

Example change to ABNF for Runtime Expressions:
```
name = 1*CHAR ; Ensures 'name' is at least one CHAR
```

In a related question, the ABNF is silent on allowed characters in CHAR. Should that be defined for clarity? Are carriage returns and line feeds allowed? In my thoughts, they wouldn't be allowed.

char0n:
This change will also allow to normalize `token` during the parsing.

Closes #253

### DIFF
--- a/src/arazzo.md
+++ b/src/arazzo.md
@@ -845,6 +845,10 @@ The runtime expression is defined by the following [ABNF](https://tools.ietf.org
         "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
 ```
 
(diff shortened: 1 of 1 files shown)


char0n:
Refs https://github.com/OAI/OpenAPI-Specification/pull/4368

ndenny (approved):
Yup, matches the OpenAPI spec language 👍

Issue #270, opened 2024-11-07
Feature: Add Support for AsyncAPI v2/v3 as Source Definition


ivangsa:
Adding support for AsyncAPI definitions would extend the types of workflows that can be described with Arazzo, as many organizations are now combining Request/Response APIs (OpenAPI) with Event-Driven architectures (AsyncAPI).

The latest version of AsyncAPI is v3, but since v2 is still the most widely adopted version, the intention is to provide support for both versions as feasible, and prioritize v3 only when supporting both becomes impractical.

## Required Changes to Arazzo Specification

- Add `asyncapi` as a source definition type.
- Extend the meaning of `workflow.step.operationId` to include pointing to an AsyncAPI `operationId` (either v2 or v3):
  - [AsyncAPI v2 operationId](https://v2.asyncapi.com/docs/reference/specification/v2.5.0#operationObject)
  - [AsyncAPI v3 operationId](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationsObject)
- Extend the meaning of `workflow.step.parameters.in`:
  - `header` should also include AsyncAPI message headers.
  - `path` should include AsyncAPI `channel.parameters`, as they are similar to OpenAPI `path` parameters (or consider adding a `channel` value for disambiguation).
- Runtime expressions already include support for `$message.body` and `$message.header` this would be used for context in successCriterias 

## Additional Considerations for **Broker-Intermediated APIs**

While AsyncAPI and OpenAPI are similar, and it’s tempting to use the `operationId` in AsyncAPI, there is a key difference: broker-mediated APIs are reciprocal (publishing/sending on one side is subscribing/receiving on the other).
   - Sometimes, an API might only document one side’s perspective. For example:
     - A service that publishes events might document its "publish" operations.
     - However, it might not document the "subscribe" operation that consumers would use to receive these events (consumers just use the "publish" operation documentation and tooling).
   - Teams and organizations may model this symmetry differently, and sometimes the "inverse" operation may not be defined at all.

To support this, we would need an alternative way to reference the **reciprocal operation**.

## Additional Changes to Arazzo Specification

When the desired `operationId` is not documented in the existing AsyncAPI, we can support referencing the `channel` instead of the operation and specify the reciprocal action (`send|receive`).

- Add `workflow.step.channel` as a fixed field pointing to an AsyncAPI channel:
  - [AsyncAPI v2 channels](https://v2.asyncapi.com/docs/reference/specification/v2.5.0#channelsObject)
[...]

PR #367, opened 2025-09-09
feat: added support for AsyncAPI v3

nashjain:
- Added support for AsyncAPI v3
- Also cleaned up the spec to make it very clear that step-object can be oneOf openapi-step-object or asyncapi-step-object or workflow-step-object 
- For AsyncAPI we really need support for timeout, fork and join. However, these are also useful for OpenAPI so added it at the base step object. 
- For OpenAPI we need at least one successCriteria but for AsyncAPI it can be optional.

While we've tried to incorporate as much as possible from #270 not everything is covered.


frankkilcommins:
Thanks @nashjain - I will try to propose changes to the specification based on the examples later this week. There will no doubt be a few finer grained points that we'll need to discuss

nashjain:
Thanks @frankkilcommins The proposal mostly looks good to me. I just need sometime to think through a couple of items. I'm currently in Australia. Next week, once I'm back we could jump on a call to discuss and close it.

src/arazzo.md
```diff
@@ -173,11 +178,35 @@ workflows:
...
+    onSuccess:
+      - name: 'noPetsAvailable'
+        type: "end"
+        criteria:
+          - condition: $response.body#/0 == null
     outputs:
-      # outputs from this step
-      availablePets: $response.body
+      petId: $response.body#/0/id
+  - stepId: purchasePetStep
```
DmitryAnansky, line 281: Could you please update the operationId in the first step ` operationId: loginUser`.
```
If multiple (non arazzo type) sourceDescriptions are defined, then the operationId MUST be specified using a [Runtime Expression](https://spec.openapis.org/arazzo/latest.html#runtime-expressions) (e.g., $sourceDescriptions.<name>.<operationId>) to avoid ambiguity or potential clashes.
```

src/arazzo.md
```diff
@@ -296,19 +325,24 @@ Describes a single workflow step which MAY be a call to an API operation ([OpenA
...
+| <a name="stepDependsOn"></a>dependsOn             |                                      List[`string`]                                      | A list of steps that MUST be completed before this step can be executed. Each value provided MUST be a `stepdId`. The keys must follow the regular expression: `^\$steps.<stepId>$`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
```
DmitryAnansky: Does this mean that Step Object DependsOn can be only defined with the step inside `current` workflow?

src/arazzo.md
```diff
@@ -377,6 +411,32 @@ steps:
...
+        in: header
+        value: $inputs.correlationId
+  requestBody:
+      payload:
+          productId: $inputs.productDetails.productId
+          quantity: $inputs.productDetails.quantity
+- stepId: confirmOrder
+  description: This step demonstrates the action of receiving a message payload to confirm an order
+  operationId: $sourceDescriptions.asyncOrderApi.confirmOrder
+  correlationId: $steps.placeOrder.outputs.requestId
```
DmitryAnansky: Could you please define outputs in `stepId: placeOrder` before using it in `stepId: confirmOrder`

src/arazzo.md
```diff
@@ -296,19 +325,24 @@ Describes a single workflow step which MAY be a call to an API operation ([OpenA
...
+| <a name="stepOnFailure"></a>onFailure             | [[Failure Action Object](#failure-action-object) \| [Reusable Object](#reusable-object)] | An array of failure action objects that specify what to do upon step failure. If omitted, the default behavior is to break and return. If multiple failure actions have similar `criteria`, the first sequential action matching the criteria SHALL be the action executed. If a failure action is already defined at the [Workflow](#workflow-object), the new definition will override it but can never remove it. If a Reusable Object is provided, it MUST link to a failure action defined in the [components](#components-object) of the current Arazzo document. The list MUST NOT include duplicate failure actions.                        |
```
DmitryAnansky: I am wondering if you already discussed how asyncApi steps should behave with `retry` on Failure actions.
Will it subscribe again, or do any other actions? How to handle potentially duplicated messages in queues?
Maybe I am missing smth.

src/arazzo.md
nashjain: Thanks. This has been fixed now.

src/arazzo.md
nashjain: This was not explicit. Now we've updated the pattern and description to make it explicit that you can depend on steps in the same workflow or steps in other workflow, in which case you need to provide the fully qualified stepId with the workflow id.

src/arazzo.md
nashjain: Thanks. This has been fixed now. We are directly ref to the input and we don't need to depend on outputs.

src/arazzo.md
nashjain: That's a great question. I don't think we've have explicitly discussed this. Here is my thinking:
1. If a `receive` async step fails before an ack, retry would just try to fetch the message again from the topic/queue again. It is safe to retry.
2. If a `send` async step fails, it could mean 
    1. the message was never sent (in which case it is safe to retry) OR 
    2. message was sent, but we are not sure what happened. However since we recommend sending a correlationId, even if a duplicate message was sent, the consumer can ignore it
  
Most messaging systems focus on delivery and ordering guarantees, not uniqueness. It is expected that the consumer uses the correlationId to check for duplicates.

Do you think Arazzo should do something special in this case?

frankkilcommins: @nashjain what you state in your thinking seems reasonable to me.

frankkilcommins (commented):
@nashjain can you please update the PR to target `v1.1-dev` branch? Thanks

src/arazzo.md
DmitryAnansky: @frankkilcommins / @nashjain 
I am missing clear vision on how a mix of OpenAPI and AsyncAPI steps should work.
Specifically the time when `outputs` set even with the usage of `dependsON`.
Could you please help me understand future flow for tooling?

Lets imagine the structure

```
Workflow:
steps:
- step1 (Async)
- step2 (Sync) + dependsOn(step1) - to outputs to be set
- step3 (Async)
- step4 (Sync)
```
In current scenario:
1) should we pause execution on step2, as it dependsOn step1?
2) should step3 be executed after step2 or step1 ?
3) what if step1 can receive not only one response ?

nashjain: @DmitryAnansky here is a working example I demonstrated at the OAI Conference in Paris last week. Also attached it the sequence diagram for this flow. Please go through this example and let me know if this makes sense?

```yaml
arazzo: "1.0.0"
info:
  title: "Arazzo Workflow"
  version: "1.0.0"
sourceDescriptions:
- name: "LocationApi"
  url: "./openapi/location.yaml"
  type: "openapi"
- name: "ProductApi"
  url: "./openapi/product.yaml"
  type: "openapi"
- name: "AsyncOrderApi"
  url: "./asyncapi/order.yaml"
  type: "asyncapi"
- name: "WarehouseApi"
  url: "./openapi/warehouse.yaml"
  type: "openapi"
- name: "OrderApi"
  url: "./openapi/order.yaml"
  type: "openapi"
workflows:
- workflowId: "PlaceOrder"
  inputs:
    required:
    - "createOrderSend"
    - "getUserLocation"
    type: "object"
    properties:
      createOrderSend:
        required:
        - "requestId"
        type: "object"
        properties:
          requestId:
            type: "string"
      getUserLocation:
        required:
        - "userEmail"
        type: "object"
        properties:
          userEmail:
            type: "string"
            format: "email"
  steps:
  - stepId: "getUserLocation"
    operationId: "$sourceDescriptions.LocationApi.getUserLocation"
    parameters:
    - name: "userEmail"
      in: "query"
      value: "$inputs.getUserLocation.userEmail"
    successCriteria:
    - condition: "$statusCode == 200"
    outputs:
      userId: "$response.body#/userId"
      locationCode: "$response.body#/locationCode"
  - stepId: "getProducts"
    operationId: "$sourceDescriptions.ProductApi.getProducts"
    parameters:
    - name: "locationCode"
      in: "query"
      value: "$steps.getUserLocation.outputs.locationCode"
    successCriteria:
    - condition: "$statusCode == 200"
    onSuccess:
    - name: "IsArrayEmpty"
      type: "end"
      criteria:
      - condition: "$response.body#/0 == null"
    outputs:
      productId: "$response.body#/0/productId"
      inventory: "$response.body#/0/inventory"
  - stepId: "createOrderSend"
    operationId: "$sourceDescriptions.AsyncOrderApi.createOrder"
    action: "send"
    parameters:
    - name: "requestId"
      in: "header"
      value: "$inputs.createOrderSend.requestId"
    requestBody:
      payload:
        userId: "$steps.getUserLocation.outputs.userId"
        productId: "$steps.getProducts.outputs.productId"
        inventory: "$steps.getProducts.outputs.inventory"
    outputs:
      requestId: "$message.header.requestId"
  - stepId: "createOrderReceive"
    operationId: "$sourceDescriptions.AsyncOrderApi.createOrder"
    action: "receive"
    correlationId: "$steps.createOrderSend.outputs.requestId"
    timeout: 6000
    outputs:
      orderId: "$message.payload.orderId"
  - stepId: "reserveInventory"
    operationId: "$sourceDescriptions.WarehouseApi.reserveInventory"
    dependsOn:
    - "$steps.createOrderReceive"
    parameters:
    - name: "orderId"
      in: "query"
      value: "$steps.createOrderReceive.outputs.orderId"
    successCriteria:
    - condition: "$statusCode == 200"
  - stepId: "orderAccepted"
    operationId: "$sourceDescriptions.AsyncOrderApi.orderAccepted"
    action: "receive"
    correlationId: "$steps.createOrderSend.outputs.requestId"
    timeout: 6000
  - stepId: "outForDelivery"
    operationId: "$sourceDescriptions.AsyncOrderApi.outForDelivery"
    action: "send"
    dependsOn:
    - "$steps.createOrderReceive"
    parameters:
    - name: "requestId"
      in: "header"
      value: "$steps.createOrderSend.outputs.requestId"
    requestBody:
      payload:
        orderId: "$steps.createOrderReceive.outputs.orderId"
  - stepId: "getOrderDetails"
    operationId: "$sourceDescriptions.OrderApi.getOrderDetails"
    dependsOn:
    - "$steps.createOrderReceive"
    parameters:
    - name: "orderId"
      in: "path"
      value: "$steps.createOrderReceive.outputs.orderId"
    successCriteria:
    - condition: "$statusCode == 200"
components: {}
```
flow

DmitryAnansky: Thank you for sharing the example.
It’s a bit hard to see the whole picture without more descriptive details.
Could you please shed more light on the case related to the following part of the flow:

So, we have this synchronous step.
```yaml
  - stepId: "reserveInventory"
    operationId: "$sourceDescriptions.WarehouseApi.reserveInventory"
    dependsOn:
    - "$steps.createOrderReceive"
```
  
I can see that it `dependsOn` the `createOrderReceive` async step:

```yaml
    - stepId: "createOrderReceive"
    operationId: "$sourceDescriptions.AsyncOrderApi.createOrder"
....
```

However, it is not clear from the execution diagram when the `orderAccepted` step should be executed.
Maybe you have a diagram with stepIds? This would be very helpful.

```yaml
  - stepId: "orderAccepted"
    operationId: "$sourceDescriptions.AsyncOrderApi.orderAccepted"
....
```

If I understand it correctly, the steps in the example execute one by one, even though some of them are async?

kevinduffey (dismissed):
Overall looks fantastic. We've discussed a bunch and my only request at this stage is that we consider adding some sort of test details for future tool providers to utilize in ensuring their tool supports the addition of async.

src/arazzo.md
```diff
@@ -296,19 +325,24 @@ Describes a single workflow step which MAY be a call to an API operation ([OpenA
...
+| <a name="stepDependsOn"></a>dependsOn             |                                      List[`string`]                                      | A list of steps that MUST be completed before this step can be executed. Each value provided MUST be a `stepdId`. If you depend on a step from another workflow, you MUST use the full reference format with the workflow id. The keys MUST follow the regular expression: `^\$(?:workflows\.[^.]+\.)?steps\.[^.]+$`.                                                                                                                                                                                                                                                                                                                               |
```
tatomyr: Does introducing the step-level `dependsOn` implies parallel execution of steps? I was under the impression that steps are intended to be executed sequentially regardless whether they are rest or async.

frankkilcommins: @tatomyr - thanks for raising this. It's an important point that we could clarify within the spec.

IMO, Introducing step-level `dependsOn` does not require parallel execution, but it does enable it. 

Steps without `dependsOn` MAY be considered independent and thus MAY begin execution immediately, including in parallel, if tooling supports it.

Tools that supports only sequential execution MUST fall back to topologically sorted execution of steps, respecting all declared `dependsOn` relationships are respected.

It would be beneficial to include additional authoring guidance, something along the lines of:

> If any step defines a `dependsOn`, authors SHOULD declare dependencies for _all steps_ where execution order matters. Steps without `dependsOn` or implicit **output references** MAY be interpreted as suitable for parallel execution. Do not rely on the order of the `steps` array when using `dependsOn`.

The alternative to this would be to constrain `dependsOn` usage to asynchronous operations (akin to an `await` semantics). While that simplifies the model, it limits the utility of `dependsOn` for broader use cases that would benefit from more general dependency-driven execution, including parallelism and fan-in/fan-out coordination.

tatomyr: @frankkilcommins wouldn't this be a breaking change? 

>  dependsOn does not require parallel execution, but it does enable it.

If we allow that, it can break existing workflows, as it introduces ambiguity in step sequencing that don't have the `dependsOn` relation. E.g., a workflow describing some CRUD sequence can potentially be executed simultaneously (unless `dependsOn` is explicitly added for each step), which might not be what the user expects.

Also, this looks to me overly complicated and error-prone, because it becomes harder to write, easier to miss something, and way harder to read, as now you'll have to do mental gymnastics to re-sort the steps. 
Even in [the example provided](https://github.com/OAI/Arazzo-Specification/pull/367#discussion_r2627978295), the steps `createOrderReceive` and `createOrderSend` could potentially be executed in parallel (because there's no `dependsOn` declared between them), but it doesn't make much sense since the first does depend on the  latter (so I guess it was supposed to have the `dependsOn` relation, but that was simply missed).

Again, what is the need for steps parallelism for AsyncAPI operations in the first place? If we do want parallelism, it can be achieved via different workflows.

frankkilcommins: @tatomyr thanks for the response.

My comment was mainly highlighting that if we introduce `dependsOn` at the step level, then it's reasonable to expect similar semantics to `dependsOn` at the workflow level, where parallel execution is permitted (though not required and it's tooling-dependent). This potential wasn't something explicitly discussed when adding AsyncAPI support, so it definitely warrants further review. We should also re-evaluate whether `timeout` (or another mechanism) might already cover the core needs for async steps, without introducing additional complexity.

To your points:

> wouldn't this be a breaking change?

Not necessarily, it depends on how we define the behaviour in the spec. Existing workflows that do not use `dependsOn` at the step level would continue to execute sequentially, as they always have. So there’s no break in behaviour unless authors opt into the new capability.

> Also, this looks to me overly complicated and error-prone, because it becomes harder to write, easier to miss something, and way harder to read, as now you'll have to do mental gymnastics to re-sort the steps.

Yes, dependency-driven execution adds some complexity, but it’s an opt-in feature. Authors would only use step-level `dependsOn` when they want more control, such as when mixing asynchronous operations or orchestrating parallel execution. For most workflows, the default sequential model remains simpler and unchanged.

> Steps like createOrderReceive and createOrderSend could now be executed in parallel

Yes, that’s a valid observation. The current example and PR may not fully account for how `dependsOn` will be interpreted with respect to parallelism. If we go forward with step-level `dependsOn`, we need to tighten both examples and authoring guidance to reflect the intended semantics clearly.

> What is the need for step parallelism for AsyncAPI operations

There's no explicit need per se, but the possibility of parallelism naturally arises from the semantics of `dependsOn`, given how it behaves at workflow level. We shouldn’t assume this feature is only for AsyncAPI, it opens up broader orchestration patterns, which some use cases may benefit from.

So in summary, we need to review whether `dependsOn` at the step level is the right mechanism to solve the “await” behaviour needed for AsyncAPI support. If we proceed with it, the spec should clearly define its semantics, especially around sequencing and parallelism.

There will likely be a desire to align its behaviour with workflow-level `dependsOn`, but we can define clear boundaries and expectations if needed.

That said, it’s also worth exploring whether existing mechanisms like `timeout` (or a more focused addition), might already address the async coordination need, without adding new complexity to the step model.

tatomyr: @frankkilcommins thanks for the clarification.

> If we proceed with it, the spec should clearly define its semantics

Yes, this is what I'd expect, along with an example demonstrating that. 

> it’s also worth exploring whether existing mechanisms like timeout (or a more focused addition), might already address the async coordination need

From the same example I've surmised that the `correlationId`/`timeout` do the job of coordinating the async calls, unless I'm missing something.

frankkilcommins: @nashjain one last push needed here to get this over the line, but the points above are valid. Do you want to determine if the combination of `correlationId`, `timeout`, and the context that the operation is within an `asyncapi` type sourceDescription are enough for us to coordinate the async behaviour?

Or if we want to define very explicit semantics about a step-level `dependsOn`?

kevinduffey (dismissed):
LGTM

spirefyio (approved):
Get it done!

kevinduffey (dismissed):
LGTM

src/schemas/validation/schema.yaml
```diff
@@ -179,13 +179,13 @@ $defs:
         description: The duration in milliseconds to wait before timing out the step
         type: integer
       dependsOn:
-        description: Specifies a list of step identifiers that must complete (or be waited for) before the current step can begin execution. If you depend on a step from another workflow, you MUST use the full reference format with the workflow id.
+        description: Specifies a list of step identifiers that must complete (or be waited for) before the current step can begin execution. Steps referred by dependsOn SHOULD be non-blocking/async steps. If you depend on a step from another workflow, you MUST use the full reference format with the workflow id.
```
DmitryAnansky: Could you please clarify, why step mentioned in dependsOn can't be synchronous? 
`Steps referred by dependsOn SHOULD be non-blocking/async steps.`

DmitryAnansky: Like can async Step depend on Sync step results?

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+  - stepId: "orderAccepted"
+    operationId: "$sourceDescriptions.AsyncOrderApi.orderAccepted"
+    action: "receive"
+    correlationId: "$steps.createOrderSend.outputs.requestId"
+    timeout: 6000
+  - stepId: "outForDelivery"
+    operationId: "$sourceDescriptions.AsyncOrderApi.outForDelivery"
+    action: "send"
+    dependsOn:
+    - "$steps.createOrderReceive"
```
DmitryAnansky: Do we really need to reference `$steps.createOrderReceive`?
Since stepId is already unique, maybe it would make sense to use just `createOrderReceive`.

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+    operationId: "$sourceDescriptions.AsyncOrderApi.createOrder"
+    action: "receive"
+    correlationId: "$steps.createOrderSend.outputs.requestId"
+    timeout: 6000
+    outputs:
+      orderId: "$message.payload.orderId"
+  - stepId: "reserveInventory"
+    operationId: "$sourceDescriptions.WarehouseApi.reserveInventory"
+    dependsOn:
+    - "$steps.createOrderReceive"
```
DmitryAnansky: Can it be a step from another workflow?

kevinduffey: yes

DmitryAnansky: Could you please also add an example with the usage?
Will it be like `$workflows.{workflowId}.steps.{stepId}` or from external file,
`$sourceDescriptions.{name}.workflows.{workflowId}.steps.{stepId}`

src/schemas/validation/schema.yaml
nashjain: Technically the spec does not stop you. But the intent is for a step (sync or async) to use the output of an asynchronous step. Hence we've used the language around SHOULD and not MUST.

With an async step, once you invoke it, you would move forward, as it is a non-blocking call. However, with a sync step, since it is blocking, you would anyway need to wait for it to complete.

Is there a use case you've in mind for which you want to depend on a sync step?

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
nashjain: These examples are unit tests purely for testing purpose, not user documentation. We will be doing that separately.

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
nashjain: This is existing syntax. Nothing to do with this PR.

DmitryAnansky: But it is not existing syntax.
Previously `dependsOn` property was only on workflows level, and syntax is like `$sourceDescriptions.<name>.<workflowId>`

```
A list of workflows that MUST be completed before this workflow can be processed. Each value provided MUST be a workflowId. If the workflow depended on is defined within the current Workflow Document, then specify the workflowId of the relevant local workflow. If the workflow is defined in a separate Arazzo Document then the workflow MUST be defined in the sourceDescriptions and the workflowId MUST be specified using a [Runtime Expression](https://spec.openapis.org/arazzo/latest.html#runtime-expressions) (e.g., $sourceDescriptions.<name>.<workflowId>) to avoid ambiguity or potential clashes.
```

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
DmitryAnansky: It would be nice to have this case covered in unit tests as well.

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
nashjain: If you see, $steps.\<stepId\>.outputs.\<outputName\> is widely used. So I simply picked $steps.\<stepId\> from there. That's what I meant by existing syntax.

DmitryAnansky: @nashjain 
In your example `"$steps.createOrderReceive"` will evaluate to an object according to the existing specification of [Runtime Expressions.](https://spec.openapis.org/arazzo/latest.html#runtime-expressions)
e.g. =>
```
$steps:
   stepId:
       response:
           .....
       outputs:
           someKey1: someValue1
           ......
```
Could you please clarify what does it even mean to dependsOn an object?

src/schemas/validation/schema.yaml
tatomyr: I’m wondering in what case we would need to make the dependency explicit. It seems redundant to me, since each step has to be executed (including setting the outputs) before we proceed to another step. Otherwise it will create the possibility to execute steps in different order (e.g., set the dependency to a next step).

Also, what does '... step identifiers that must complete (or be waited for)'? actually mean? How do we determine that an async step has completed? What exactly are we waiting for?

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
frankkilcommins: This is likely due to how we describe what to do for workflow `dependsOn`:
> A list of workflows that MUST be completed before this workflow can be processed. Each value provided MUST be a workflowId. If the workflow depended on is defined within the current Workflow Document, then specify the workflowId of the relevant local workflow. If the workflow is defined in a separate Arazzo Document then the workflow MUST be defined in the sourceDescriptions and the workflowId MUST be specified using a [Runtime Expression](https://github.com/specmatic/Arazzo-Specification/blob/dev/src/arazzo.md#runtime-expressions) (e.g., `$sourceDescriptions.<name>.<workflowId>`) to avoid ambiguity or potential clashes.

I think the intent is fairly clear but in general we should improve the ABNF grammar. Time permitting, I'll create a PR for #426 as part of the 1.1 efforts

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
frankkilcommins: Related to this: As part of the 1.1 release, I am hoping that we can add an Arazzo section to the OAI learn site and have lots of good rich examples there.

frankkilcommins: > Will it be like `$workflows.{workflowId}.steps.{stepId}` or from external file,
`$sourceDescriptions.{name}.workflows.{workflowId}.steps.{stepId}`

- `$workflows.{workflowId}.steps.{stepId}` is valid if the workflow is defined in the local Arazzo Document.
- `$sourceDescriptions.{name}.workflows.{workflowId}.steps.{stepId}` is valid if the workflow is defined in and external Arazzo Document

DmitryAnansky: @frankkilcommins 
The following syntax appears clear and consistent:
`$sourceDescriptions.{name}.workflows.{workflowId}.steps.{stepId}`
However, this alternative form raises some questions:
`$workflows.{workflowId}.steps.{stepId}`

 It is unclear why the `$workflows.{workflowId}.steps.{stepId}` does not have same context as `$steps.{stepId}`.
 Please correct me if I am wrong, but the expectations seems to be that this should have similar object structure:
 `$workflows.{workflowId}.steps.{stepId}`
 ```
 outputs: {...},
 request: {...},
 response: {...},
 ```

frankkilcommins: @DmitryAnansky 

Yes `$sourceDescriptions.{name}.workflows.{workflowId}.steps.{stepId}` is the mechanism to refer to a step that's located within a remote Arazzo Document.

I believe what @nashjain was laying out in the example, was a means to refer to a step located in another workflow within the same Arazzo Document.

  - `$steps.{stepId}` operates in the _current workflow context_ and references steps within the currently executing workflow
  - `$workflows.{workflowId}.steps.{stepId}` requires _explicit workflow context_. And while you _could_ refer to the current workflow being executed, you most likely would only do this when the step is from another  workflows in the same document

  This is similar to how `$inputs.username` references the current workflow's inputs, while `$workflows.foo.inputs.username` explicitly references another workflow's inputs.
  
We need to improve the clarity of the ABNF expressions. We'll address that as part of #426 and #427 as part of v1.1 work, rather than trying to solve in this PR.

src/arazzo.md
```diff
@@ -298,19 +327,24 @@ Describes a single workflow step which MAY be a call to an API operation ([OpenA
...
+| <a name="stepDependsOn"></a>dependsOn             |                                      List[`string`]                                      | A list of steps that MUST be completed before this step can be executed. Each value provided MUST be a `stepdId`. If you depend on a step from another workflow, you MUST use the full reference format with the workflow id. The keys MUST follow the regular expression: `^\$(?:workflows\.[^.]+\.)?steps\.[^.]+$`. If your step depends on the output of a non-blocking/async step, then you SHOULD use dependsOn in your current step and reffer to the async step                                                                                                                                                                              |
```
frankkilcommins:
```suggestion
| <a name="stepDependsOn"></a>dependsOn | List[`string`] | A list of steps that MUST be completed before this step can be executed. Each value provided MUST be a `stepId`. The `stepId` value is case-sensitive. If the step depended on is defined within the **current workflow**, specify the `stepId` directly (e.g., `authStep`). If the step is defined in a **different workflow within the current Arazzo Document**, reference it using `$workflows.<workflowId>.steps.<stepId>`. If the step is defined in a **separate Arazzo Document**, the workflow MUST be defined in `sourceDescriptions` and referenced using `$sourceDescriptions.<name>.<workflowId>.steps.<stepId>` to avoid ambiguity. If your step depends on the output of a non-blocking/asynchronous step, then you SHOULD use `dependsOn` and refer to the async step using one of these patterns. |
```

There's some typos in the current version, and I think it would be good to lay out the three types of references (local workflow step, step in another workflow in same doc, step in remote doc)

frankkilcommins (changes requested):

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
nashjain: I updated the schema to clearly call out the 3 possibilities for dependsOn:

```yaml
dependsOn:
  description: Specifies a list of step identifiers that must complete (or be waited for) before the current step can begin execution. Steps referred by dependsOn SHOULD be non-blocking/async steps. Steps in the current workflow MUST be referenced directly by stepId. Steps in another workflow in the same Arazzo document MUST use $workflows.<workflowId>.steps.<stepId>. Steps in another Arazzo document MUST use $sourceDescriptions.<name>.<workflowId>.steps.<stepId>.
  type: array
  uniqueItems: true
  minItems: 1
  items:
    oneOf:
      - type: string
        pattern: ^(?!\$).+$
      - type: string
        pattern: ^\$workflows\.[^.]+\.steps\.[^.]+$
      - type: string
        pattern: ^\$sourceDescriptions\.[^.]+\.[^.]+\.steps\.[^.]+$
```

nashjain: For local reference, I dropped $steps. from the $steps.{stepId}, now we can directly refer to the stepId name. This is now consistent with other places.

tests/schema/pass/async-depends-on-across-document.arazzo.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - stepId: useLocalDirectStepId
+        operationId: localSend
+        action: send
+        dependsOn:
+          - waitForLocalEvent
+      - stepId: useExternalWorkflowStep
+        operationId: localSend
+        action: send
+        dependsOn:
+          - "$sourceDescriptions.otherWorkflowDoc.remoteWorkflow.steps.remoteReceive"
```
DmitryAnansky: Why it is string?
Should it be like 
```
dependsOn:
    -  $sourceDescriptions.otherWorkflowDoc.remoteWorkflow.steps.remoteReceive
```

tests/schema/pass/async-depends-on-cross-workflow.arazzo.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - stepId: produceEvent
+        operationId: createUser
+        action: send
+      - stepId: waitForEvent
+        operationId: fetchUserId
+        action: receive
+        correlationId: id123
+        timeout: 6000
+        outputs:
+          userId: "$message.payload.userId"
```
DmitryAnansky:
```suggestion
          userId: $message.payload.userId
```

tests/schema/pass/async-depends-on-cross-workflow.arazzo.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - workflowId: consumerWorkflow
+    steps:
+      - stepId: processEvent
+        operationId: updateUser
+        action: send
+        dependsOn:
+          - "$workflows.producerWorkflow.steps.waitForEvent"
```
DmitryAnansky:
```suggestion
          - $workflows.producerWorkflow.steps.waitForEvent
```

tests/schema/pass/async-depends-on-cross-workflow.arazzo.yaml
```diff
@@ -0,0 +1,32 @@
...
+        parameters:
+          - name: userId
+            in: header
+            value: "$workflows.producerWorkflow.steps.waitForEvent.outputs.userId"
```
DmitryAnansky:
```suggestion
            value: $workflows.producerWorkflow.steps.waitForEvent.outputs.userId
```

tests/schema/pass/async-depends-on.arazzo.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - stepId: step1
+        operationId: createUser
+        action: send
+      - stepId: step2
+        operationId: fetchUserId
+        action: receive
+        correlationId: id123
+        timeout: 6000
+        outputs:
+          userId: "$message.payload.userId"
```
DmitryAnansky:
```suggestion
          userId: $message.payload.userId
```

tests/schema/pass/async-depends-on.arazzo.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - stepId: step3
+        operationId: updateUser
+        action: send
+        dependsOn:
+          - "step2"
+        parameters:
+          - name: userId
+            in: header
+            value: "$steps.step2.outputs.userId"
```
DmitryAnansky:
```suggestion
            value: $steps.step2.outputs.userId
```

tests/schema/pass/async-depends-on.arazzo.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - stepId: step4
+        operationId: deleteUser
+        action: send
+        dependsOn:
+          - "$workflows.dependsOnWorkflow.steps.step2"
```
DmitryAnansky:
```suggestion
          - $workflows.dependsOnWorkflow.steps.step2
```

tests/schema/pass/async-depends-on.arazzo.yaml
```diff
@@ -0,0 +1,39 @@
...
+        parameters:
+          - name: userId
+            in: header
+            value: "$steps.step2.outputs.userId"
```
DmitryAnansky:
```suggestion
            value: $steps.step2.outputs.userId
```

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+          userEmail:
+            type: "string"
+            format: "email"
+  steps:
+  - stepId: "getUserLocation"
+    operationId: "$sourceDescriptions.LocationApi.getUserLocation"
+    parameters:
+    - name: "userEmail"
+      in: "query"
+      value: "$inputs.getUserLocation.userEmail"
```
DmitryAnansky:
```suggestion
      value: $inputs.getUserLocation.userEmail
```

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+        required:
+        - "userEmail"
+        type: "object"
+        properties:
+          userEmail:
+            type: "string"
+            format: "email"
+  steps:
+  - stepId: "getUserLocation"
+    operationId: "$sourceDescriptions.LocationApi.getUserLocation"
```
DmitryAnansky:
```suggestion
    operationId: $sourceDescriptions.LocationApi.getUserLocation
```

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+    parameters:
+    - name: "userEmail"
+      in: "query"
+      value: "$inputs.getUserLocation.userEmail"
+    successCriteria:
+    - condition: "$statusCode == 200"
+    outputs:
+      userId: "$response.body#/userId"
```
DmitryAnansky:
```suggestion
      userId: $response.body#/userId
```

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+            format: "email"
+  steps:
+  - stepId: "getUserLocation"
+    operationId: "$sourceDescriptions.LocationApi.getUserLocation"
+    parameters:
+    - name: "userEmail"
+      in: "query"
+      value: "$inputs.getUserLocation.userEmail"
+    successCriteria:
+    - condition: "$statusCode == 200"
```
DmitryAnansky:
```suggestion
    - condition: $statusCode == 200
```

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+      value: "$steps.createOrderReceive.outputs.orderId"
+    successCriteria:
+    - condition: "$statusCode == 200"
+  - stepId: "orderAccepted"
+    operationId: "$sourceDescriptions.AsyncOrderApi.orderAccepted"
+    action: "receive"
+    correlationId: "$steps.createOrderSend.outputs.requestId"
+    timeout: 6000
+  - stepId: "outForDelivery"
+    operationId: "$sourceDescriptions.AsyncOrderApi.outForDelivery"
```
DmitryAnansky: Is there a specific reason for wrapping the runtime expression in ""?

Looking at the example in the Arazzo specification
https://github.com/OAI/Arazzo-Specification/blob/main/examples/1.0.0/pet-coupons.arazzo.yaml and the Arazzo documentation, I don’t see a requirement to quote the expression each time it’s used.

nashjain: Thanks for catching this. I've fixed it.

src/schemas/validation/schema.yaml
tatomyr: Any updates?

tests/schema/pass/async-depends-on-across-document.arazzo.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - stepId: waitForLocalEvent
+        operationId: localReceive
+        action: receive
+        correlationId: id123
+        timeout: 6000
+      - stepId: useLocalDirectStepId
+        operationId: localSend
+        action: send
+        dependsOn:
+          - waitForLocalEvent
```
tatomyr, line 24: Is it possible to reference a workflow instead of a step?

tests/schema/pass/async-depends-on-cross-workflow.arazzo.yaml
```diff
@@ -0,0 +1,32 @@
...
+        timeout: 6000
+        outputs:
+          userId: $message.payload.userId
+  - workflowId: consumerWorkflow
+    steps:
+      - stepId: processEvent
+        operationId: updateUser
+        action: send
+        dependsOn:
+          - $workflows.producerWorkflow.steps.waitForEvent
```
tatomyr, line 28: Why not just wait for producerWorkflow?

frankkilcommins (dismissed):
Thanks for all the work on this @nashjain. We might tidy up examples to remove redundant quotes etc. but that can happen in a small house-keeping PR as we prepare information for the learn site.

src/arazzo.md
```diff
@@ -379,6 +413,32 @@ steps:
...
+  requestBody:
+      payload:
+          productId: $inputs.productDetails.productId
+          quantity: $inputs.productDetails.quantity
+- stepId: confirmOrder
+  description: This step demonstrates the action of receiving a message payload to confirm an order
+  operationId: $sourceDescriptions.asyncOrderApi.confirmOrder
+  correlationId: $inputs.correlationId
+  action: receive
+  dependsOn: placeOrder
```
RomanHotsiy: This example is wrong. The spec defines the `dependsOn` as array of strings

RomanHotsiy: there are more examples like this throughout this PR

src/arazzo.md
```diff
@@ -298,19 +327,24 @@ Describes a single workflow step which MAY be a call to an API operation ([OpenA
...
+| <a name="stepDependsOn"></a>dependsOn             |                                      List[`string`]                                      | A list of steps that MUST be completed before this step can be executed. Each value provided MUST be a `stepId`. The `stepId` value is case-sensitive. If the step depended on is defined within the **current workflow**, specify the `stepId` directly (e.g., `authStep`). If the step is defined in a **different workflow within the current Arazzo Document**, reference it using `$workflows.<workflowId>.steps.<stepId>`. If the step is defined in a **separate Arazzo Document**, the workflow MUST be defined in `sourceDescriptions` and referenced using `$sourceDescriptions.<name>.<workflowId>.steps.<stepId>` to avoid ambiguity. If your step depends on the output of a non-blocking/asynchronous step, then you SHOULD use `dependsOn` and refer to the async step using one of these patterns. |
```
RomanHotsiy: There is no definition of what is "non-blocking/asynchronous step" anywhere in the spec.
What is this?

RomanHotsiy (changes requested):

src/arazzo.md
RomanHotsiy: I would also expand more on the semantics. It's not clear from the spec if the dependency should trigger step execution. I guess it should not and the step should be executed in some other way and dependsOn is just waiting on the result. But It would benefit the spec  clarity if we add this note.

src/arazzo.md
nashjain: Thanks I fixed the dependsOn: placeOrder. I could not find any other example. Can you please highlight where else did you spot such issues?

tests/schema/pass/async-depends-on.arazzo.yaml
```diff
@@ -0,0 +1,39 @@
...
+workflows:
+  - workflowId: dependsOnWorkflow
+    steps:
+      - stepId: step1
+        operationId: createUser
+        action: send
+      - stepId: step2
+        operationId: fetchUserId
+        action: receive
+        correlationId: id123
```
AlbinaBlazhko17, line 18: Shouldn't the `correlationId` on a `receive` step always reference an output from the preceding send step, rather than being a hardcoded value?

tests/schema/pass/async-depends-on.arazzo.yaml
```diff
@@ -0,0 +1,39 @@
...
+          - step2
+        parameters:
+          - name: userId
+            in: header
+            value: $steps.step2.outputs.userId
+      - stepId: step4
+        operationId: deleteUser
+        action: send
+        dependsOn:
+          - $workflows.dependsOnWorkflow.steps.step2
```
AlbinaBlazhko17: Regarding the spec:
> Steps in the current workflow MUST be referenced directly by stepId.

It should just be `step2`.

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - name: orderId
+      in: query
+      value: $steps.createOrderReceive.outputs.orderId
+    successCriteria:
+    - condition: $statusCode == 200
+  - stepId: orderAccepted
+    operationId: $sourceDescriptions.AsyncOrderApi.orderAccepted
+    action: receive
+    correlationId: $steps.createOrderSend.outputs.requestId
+    timeout: 6000
```
AlbinaBlazhko17, line 107: There are no success criteria. How can we be sure that this step has been executed successfully?

src/schemas/validation/schema.yaml
```diff
@@ -240,21 +263,94 @@ $defs:
...
+          action:
+            description: Specifies the intended operation on the async channel, indicating whether the action is sending data to the channel or receiving data from the channel
+            enum:
+              - send
+              - receive
+          successCriteria:
+            description: A list of assertions to determine the success of the step
+            type: array
+            uniqueItems: true
+            minItems: 0
```
AlbinaBlazhko17: Could you please explain the difference between the `OpenAPI` successCriteria object and the `AsyncAPI`?
This schema allows the user to declare `successCriteria` as an empty array.
Could you also provide an example of how we can use the `successCriteria` object with asynchronous operations?

nashjain: If I'm firing a notification (which is typically a fire-n-forget), can you please help me understand what kind of things you would check in the successCriteria?

It would really help if you can share a concrete real-world example, so we can resolve this question.

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
nashjain: See the below comment.

tests/schema/pass/async-depends-on.arazzo.yaml
nashjain: step2 is from a different workflow and hence you need to fully qualify step2

AlbinaBlazhko17: But there are no other workflows in this file. `step4` is inside the `dependsOnWorkflow` workflow, in the same way that `step3` depends on just `step2`.

src/schemas/validation/schema.yaml
```diff
@@ -217,14 +233,21 @@ $defs:
...
-          - workflowId
+  openapi-step-object:
     allOf:
+      - $ref: '#/$defs/step-object-base'
+      - type: object
+        properties:
+          operationPath:
+            description: A reference to a Source combined with a JSON Pointer to reference an operation
+            type: string
+          successCriteria:
```
AlbinaBlazhko17: There is a duplication, because `step-object-base` has the same [successCriteria declared](https://github.com/OAI/Arazzo-Specification/pull/367/changes#diff-1e714ffe4a88e9eada463c25ca80ec1f0474ae43bbebc753bc31c4a240533a2bR201).

src/schemas/validation/schema.yaml
```diff
@@ -240,21 +256,101 @@ $defs:
...
           properties:
             parameters:
               items:
                 oneOf:
-                  - $ref: '#/$defs/parameter-object'
                   - $ref: '#/$defs/reusable-object'
+                  - $ref: '#/$defs/parameter-object'
+                    required:
+                      - in
+      - if:
```
AlbinaBlazhko17, line 295: Shouldn't this condition be inverted? Currently it reads "if `correlationId` is present, `action` must be `receive`" -  but it doesn't enforce the other direction: that `correlationId` is required when `action` is `receive`.
Or we can define separate schemas for two actions: `send` and `receive`.

src/schemas/validation/schema.yaml
nashjain: I've moved successCriteria back to step-object-base. successCriteria itself is optional, but if the key is present, then it should contain at least 1 item.

src/schemas/validation/schema.yaml
nashjain: `correlationId ` is optional as per AysncAPI spec. Hence "if correlationId is present, action must be receive"

tests/schema/pass/async-depends-on.arazzo.yaml
nashjain: Sorry I miss spoke. step2 belongs to the same workflow. Here we are trying to demonstrate that both direct reference and fully qualified reference both are valid.

tests/schema/pass/async-depends-on.arazzo.yaml
nashjain: You are right from a demonstration of how real correlationId works, correlationId should be sourced from step1 or from the real source of the id. 

However, please note, as the title suggests, the purpose of this example is to check schema validity of just dependsOn. These are internal author unit tests, not end customer training examples.

nashjain: You should refer to this example: `openapi-and-asyncapi.arazzo.yaml` that is more complete example.

tests/schema/pass/async-depends-on.arazzo.yaml
AlbinaBlazhko17: Thanks for the clarification. Should the schema be updated?
Because now [description said](https://github.com/OAI/Arazzo-Specification/pull/367/changes#diff-1e714ffe4a88e9eada463c25ca80ec1f0474ae43bbebc753bc31c4a240533a2bR182), that we MUST use only `step` in current workflow.

src/schemas/validation/schema.yaml
AlbinaBlazhko17: Got it, thanks. But Arazzo is a workflow execution and testing framework, not just a description format. In that context, a `receive` step without `correlationId` has no way to filter which incoming message belongs to the current workflow instance. In a busy channel with multiple concurrent workflow executions, the runtime would pick up the first message that arrives regardless of whether it belongs to this execution — which makes the step non-deterministic. Shouldn't `correlationId` be required on receive steps specifically within Arazzo, even if AsyncAPI itself treats it as optional?

src/schemas/validation/schema.yaml
AlbinaBlazhko17: Sure, you are right, if we talk about `fire-n-forget` which means `action: send`. But for action: `receive` the incoming message payload is available, and `successCriteria` becomes meaningful. For example, checking that the order was actually confirmed and not rejected. The example of using multiple messages:

`async.yaml`:
```yaml
channels:
 email/results:
  subscribe:
   operationId: receiveEmailResult
   message:
    oneOf:
     - $ref: '#/components/messages/EmailSuccess'
     - $ref: '#/components/messages/EmailFailure'

components:
 messages:
  EmailSuccess:
   payload:
    type: object
    properties:
     status:
      type: string
      enum: [success]
     orderId:
      type: string

  EmailFailure:
   payload:
    type: object
    properties:
     status:
      type: string
      enum: [failure]
     orderId:
      type: string
     errorCode:
      type: string
      example: SMTP_TIMEOUT
     reason:
      type: string
      example: "Recipient mailbox full"
```

`arazzo.yaml`:

```yaml
- stepId: receiveEmailResult
 operationId: $sourceDescriptions.EmailApi.receiveEmailResult
 action: receive
 correlationId: $steps.sendEmail.outputs.requestId
 timeout: 6000
 successCriteria:
  - condition: $message.payload#/status == "success"
  - condition: $message.payload#/orderId != null
```

Another approach to handling failed messages is the dead letter queue. We can handle it using `onFailure`. The `onFailure` will be fired only when timeout exceeded.

I saw your changes and example for this case. Thanks!

src/schemas/validation/schema.yaml
```diff
@@ -240,21 +256,87 @@ $defs:
...
+          - required:
+              - channelPath
+    $ref: '#/$defs/specification-extensions'
+    unevaluatedProperties: false
+  workflow-step-object:
+    allOf:
+      - $ref: '#/$defs/step-object-base'
+      - type: object
+        properties:
+          workflowId:
```
AlbinaBlazhko17, line 316: Could you please clarify this? Looking at the schema, `workflow-step-object` inherits `operationId` from `step-object-base`, but there is no constraint enforcing that `workflowId` and `operationId` are mutually exclusive. The spec said, that they are mutually exclusive, but the schema currently allows declaring both simultaneously without any validation error.

tests/schema/pass/asyncapi-success-criteria.arazzo.yaml
```diff
@@ -0,0 +1,19 @@
...
+  title: AsyncAPI step with success criteria
+  version: 1.0.0
+sourceDescriptions:
+  - name: UserAPI
+    url: https://user.com/asyncapi.yaml
+    type: asyncapi
+workflows:
+  - workflowId: asyncWorkflow
+    steps:
+      - stepId: receiveOrder
```
AlbinaBlazhko17, line 12: Thank you for the example! Could you clarify the intent of this workflow? There is no preceding `send` step that publishes a message and produces the `correlationId`. I’m not entirely sure from the example who exactly is sending the message that `receiveOrder` is waiting for?

src/schemas/validation/schema.yaml
nashjain: That's a very good catch. I've pushed a new test (`invalid-workflow-step-with-operationId.arazzo.yaml`) to prove this is working as expected.

• It works because the exclusivity is enforced indirectly by the split schema. We don't need the oneOf anymore.

  The top-level step-object is now a oneOf over three subtypes at src/schemas/validation/schema.yaml:156:

  - openapi-step-object
  - asyncapi-step-object
  - workflow-step-object

  The spec in the failing example has both operationId and workflowId. That means it cannot cleanly belong to exactly one subtype.

  For openapi-step-object, operationId is allowed and the subtype requires one of operationId or operationPath at src/schemas/validation/
  schema.yaml:236. But that subtype ends with unevaluatedProperties: false at src/schemas/validation/schema.yaml:264. Since workflowId is not part
  of the OpenAPI step shape, it counts as an unevaluated property, so the OpenAPI branch fails.

  For workflow-step-object, workflowId is required at src/schemas/validation/schema.yaml:309, but that subtype also has unevaluatedProperties: false
  at src/schemas/validation/schema.yaml:320. operationId is not part of the workflow-step shape, so that branch fails too.

  asyncapi-step-object fails immediately because it requires action at src/schemas/validation/schema.yaml:266.

  So the instance matches none of the three branches, and the outer step-object.oneOf rejects it.

  Short version: the old schema enforced exclusivity with one explicit rule; the new schema enforces it by subtype selection plus `unevaluatedProperties: false`.

tests/schema/pass/asyncapi-success-criteria.arazzo.yaml
nashjain: These are not examples, these are unit tests to validate the schema is correct.

There are many workflows in which an external system could be sending an event to your topic. So in your workflow, you would not have a send, as that is outside the scope of your workflow.

src/schemas/validation/schema.yaml
nashjain: Yes, for receive it might make sense, but successCriteria is optional so it is up to the author of the workflow. I hope the updates have clarified your question. Thanks again for asking these thoughtful questions, it is really helpful.

src/arazzo.md
nashjain: > There is no definition of what is "non-blocking/asynchronous step" anywhere in the spec. What is this?

I don’t think we need to over-specify this unless it leads to ambiguity in implementation.

“Non-blocking” / “asynchronous step” are fairly well-understood concepts in most engineering contexts, and different platforms may implement them differently anyway. Trying to define them too rigidly in the spec might actually limit flexibility.

That said, if people feel this could be interpreted inconsistently, I would request you to share a lightweight clarification or example rather than a strict definition.

Happy to hear how others think about this.

nashjain: > I would also expand more on the semantics. It's not clear from the spec if the dependency should trigger step execution. I guess it should not and the step should be executed in some other way and dependsOn is just waiting on the result. But It would benefit the spec clarity if we add this note.

Fair point. I've made the necessary changes:

_Specifies a list of step identifiers that must complete (or be waited for) before the current step can begin execution. `dependsOn` only establishes a prerequisite relationship for the current step and does not trigger execution of the referenced steps. Steps referred by dependsOn SHOULD be non-blocking/async steps. Steps in the current workflow MUST be referenced directly by stepId. Steps in another workflow in the same Arazzo document MUST use `$workflows.<workflowId>.steps.<stepId>`. Steps in another Arazzo document MUST use `$sourceDescriptions.<name>.<workflowId>.steps.<stepId>`._

tests/schema/pass/async-depends-on-across-document.arazzo.yaml
nashjain: I'm guessing you are suggesting that we should ref to a workflow id instead of a specific step inside the workflow.

My concern with referring only to the workflow id is that it makes the dependency too coarse-grained. A workflow can have multiple async steps, and another workflow may need to proceed when the first async step completes, rather than waiting for the whole workflow to finish. So if step-level cross-workflow dependencies are a valid use case, we may need a way to reference a specific async step, not just the workflow as a whole.

tests/schema/pass/async-depends-on-cross-workflow.arazzo.yaml
nashjain: Not sure I understand your question. Does my previous comment clarify the question? if not, can you please clarify your concern? 

Please note you are looking at unit tests which are meant to test the schema. This is not user documentation which teaches people will use it.

frankkilcommins (dismissed):

src/arazzo.md
RomanHotsiy: all fixed now, thx

src/arazzo.md
RomanHotsiy: > That said, if people feel this could be interpreted inconsistently, I would request you to share a lightweight clarification or example rather than a strict definition.

Well, I don't understand it 😄 . Can OpenAPI step be async/non-blocking? or only AsyncAPI step?

tests/schema/pass/async-depends-on.arazzo.yaml
AlbinaBlazhko17: Could you please update the schema?
Because now [description said](https://github.com/OAI/Arazzo-Specification/pull/367/changes#diff-1e714ffe4a88e9eada463c25ca80ec1f0474ae43bbebc753bc31c4a240533a2bR182), that we MUST use only step in current workflow.

src/schemas/validation/schema.yaml
AlbinaBlazhko17, line 316: I validated this schema and found, that we can use `operationId` and `workflowId` in the same schema simultaneously. Could you please fix this behavior?

src/schemas/validation/schema.yaml
DmitryAnansky: @frankkilcommins / @nashjain 
Also wondering how we defined `async step has completed`.
Should we wait for any message, first one or specific SuccessCriteria to pass?

tests/schema/pass/async-depends-on.arazzo.yaml
nashjain: To avoid confusion, I've dropped the fully qualified step path from this file

src/schemas/validation/schema.yaml
nashjain: Can you please share the sample? In `tests/schema/fail/invalid-workflow-step-with-operationId.arazzo.yaml` I've demonstrated that both are not possible in the same spec.

src/schemas/validation/schema.yaml
nashjain: Typically, an async dependency exists because a later step needs data from a message expected on a specific topic or channel.

So in such cases, you would depend on the output of a async receive step to complete within a timeout period. When we receive the expected message payload on the expected topic, we would consider it as complete. This is implicit and we don't need to define any SuccessCriteria.

DmitryAnansky: How do you define the “expected message payload”?
From what I understand, we can only subscribe to a channel (i.e., receive an AsyncAPI response), but the exact shape or type of the expected message isn’t always strictly enforced.

Is it possible for an error message to be received when we’re expecting a specific receive message?

I can imagine a scenario where the incoming AsyncAPI message does not match what we expect. As a result, our output might not be properly resolved. In cases where there is no successCriteria (or another explicit way to mark a step as failed), the tooling doesn’t have a clear way to determine that something went wrong.

This leads to a few issues:
- A step failure may not be detected immediately.
- Outputs could end up being undefined without any clear signal.
- The problem might only surface later, when those outputs are used in subsequent steps.

How should such situations be handled?

src/schemas/validation/schema.yaml
AlbinaBlazhko17: Yes, sure. I am using this service to validate the schema: https://www.jsonschemavalidator.net/.
I put the schema to validate, which is simplified version of what we have now:
``` json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "oneOf": [
    {
      "allOf": [
        {
          "properties": {
            "stepId": {
              "type": "string"
            },
            "operationId": {
              "type": "string"
            }
          },
          "required": [
            "stepId",
            "operationId"
          ]
        }
      ],
      "unevaluatedProperties": false
    },
    {
      "allOf": [
        {
          "properties": {
            "stepId": {
              "type": "string"
            },
            "operationId": {
              "type": "string"
            }
          },
          "required": [
            "stepId"
          ]
        },
        {
          "properties": {
            "workflowId": {
              "type": "string"
            }
          },
          "required": [
            "workflowId"
          ]
        }
      ],
      "unevaluatedProperties": false
    }
  ]
}
```

And passed to it the example:
```json
{
  "stepId": "someStep",
  "workflowId": "someWorkflow",
  "operationId": "someOperation"
}
```
With this case the example with both `workflowId` and `operationId` is valid.

frankkilcommins: Hi @nashjain, yes @AlbinaBlazhko17 is correct here in so far as the `workflow-step-object` within the JSON Schema does not check for mutual exclusivity with fields like `operationId`, `operationPath`, `channelPath`.

The sample failure Arazzo document used to test, namely https://raw.githubusercontent.com/specmatic/Arazzo-Specification/refs/heads/dev/tests/schema/fail/invalid-workflow-step-with-operationId.arazzo.yaml, is actually currently failing due to missing `sourceDescriptions`

Possibly you'd add a NOT constraint within the `workflow-step-object`

```
    not:
      anyOf:
        - required: [operationId]
        - required: [operationPath]
        - required: [channelPath]
```

nashjain: Thanks @AlbinaBlazhko17 that was a very good catch. Sorry for the oversight. I've not pushed operationId to openapi-step-object and asyncAPI-step-object to avoid this issue. This also simplified a few other things. Tests are updated accordingly and passing. /cc @frankkilcommins

frankkilcommins (dismissed):

ndenny (dismissed):

frankkilcommins:
@nashjain we've been busy merging in other approved PRs into `v1.1-dev`, and now there's some merge conflicts here for you to address.

We're happy with the PR - great work. Once the merge conflicts are address we can get it merged!

nashjain:
@frankkilcommins I've resolved the conflict. Please have a look.

examples/1.1.0/pet-asyncapi.yaml
frankkilcommins, line 1: @nashjain not sure if this file just popped in now as part of the additional merge work (or slipped through initial review), but I don't see the need to create it within the src folder. We really only want the spec md and the schemas there.

If you want to keep this file then it could go in `/examples/1.1` along with an Arazzo 1.1 Document that leverages it. Otherwise, just delete from this PR and we can handle having some 1.1 examples as part of #467

nashjain: this file was always there since it is referred from arazzo.md file. I've moved it to examples folder.

frankkilcommins (dismissed):

ndenny (dismissed):

src/schemas/validation/schema.yaml
tatomyr: > Typically, an async dependency exists because a later step needs data from a message expected on a specific topic or channel.

Thanks @nashjain. So, is the only purpose of `dependsOn` at the step level to indicate that we use outputs from a previous step (or wait for that step's side effect), rather than to control parallelism (as it does at the workflow level)? What prevents us from setting outputs in this step and using them in subsequent steps without `dependsOn` being explicitly set? Do we actually need `dependsOn` at the step level, then? My take is that we should avoid adding things to the spec that aren't strictly necessary. Moreover, introducing something similar to an existing concept but with different behaviour feels a bit confusing.

nashjain: That's a great question @tatomyr. 

If you see my original PR, we had a fork/join semantics. You would use a `fork` to allow an async step to start work that continues independently of the current linear step execution, and a later step can declare `join` to mean “do not proceed until that async work has reached its completion point.”

It was proposed that the spec already has a `dependsOn` to sever a similar purpose for workflows and we should use that instead of introducing a new keyword. We agreed and changed it to `dependsOn`

So at the step level, dependsOn is not only about consuming outputs from a previous step. In a purely sequential model, that would indeed be redundant. Its value is when an async step represents in-flight work, such as waiting for a message/callback/event, and another step needs to explicitly join on that work before continuing.

Hope this helps.

frankkilcommins: > How do you define the “expected message payload”?
> From what I understand, we can only subscribe to a channel (i.e., receive an AsyncAPI response), but the exact shape or type of the expected message isn’t always strictly enforced.
>
>Is it possible for an error message to be received when we’re expecting a specific receive message?
>
>I can imagine a scenario where the incoming AsyncAPI message does not match what we expect. As a result, our output might not be properly resolved. In cases where there is no successCriteria (or another explicit way to mark a step as failed), the tooling doesn’t have a clear way to determine that something went wrong.
>
>This leads to a few issues:
>
>- A step failure may not be detected immediately.
>- Outputs could end up being undefined without any clear signal.
>- The problem might only surface later, when those outputs are used in subsequent steps.
>
>How should such situations be handled?

@DmitryAnansky regarding your points above, they are valid concerns. I think we can address ambiguity by adding the following clarifying section to the spec:

```text
##### Defining Success for Asynchronous Steps

For steps that reference AsyncAPI operations (via `operationId` or `channelPath`), tools will send or receive messages on the specified channel as defined by the referenced AsyncAPI description. For AsyncAPI send operations, the step completes immediately after the message is sent. Arazzo does not model broker acknowledgment or delivery confirmation. For AsyncAPI receive operations, step completion depends on message receipt as described below.

Because AsyncAPI channels MAY define multiple message types, and because message payloads MAY represent either success or failure conditions, authors SHOULD define `successCriteria` for AsyncAPI receive steps to explicitly evaluate the received message payload (for example, via `$message.payload`) and determine whether the step succeeded.

Authors MAY omit `successCriteria` only when both of the following conditions are met:

- The channel defines a single message type that unambiguously represents successful completion
- The message payload does not contain fields indicating error states (e.g., status codes or error flags)

When `successCriteria` is omitted, receiving any message matching the `correlationId` (if specified) within the `timeout` period is considered a successful step completion.

If `correlationId` is specified, only messages matching the correlation identifier are considered. If no matching message is received within the `timeout` period, the step fails and triggers any defined `onFailure` actions.
```

tests/schema/pass/openapi-and-asyncapi.arazzo.yaml
frankkilcommins: ambiguity around expressions is addressed in #454 - resolving this comment thread

src/schemas/validation/schema.yaml
frankkilcommins: @tatomyr @nashjain 

Regarding comments here on step level `dependsOn` and in light of the older comment thread on the topic (see https://github.com/OAI/Arazzo-Specification/pull/367#discussion_r2671672797), I would propose the following section be added to add clarity and remove ambiguity around what step level dependsOn introduces and opens the possibility too. Whether or not tooling chooses to support all aspects is a tooling choice, but I propose some clarity in that respect too to make it easier for tooling authors to follow).

```text
##### Step Dependencies and Execution Order

The `dependsOn` field at the step level is primarily intended to coordinate asynchronous operations.

**Recommended Approach for Synchronous Workflows**

For workflows containing only synchronous steps, the RECOMMENDED approach is to order steps sequentially in the steps array without using `dependsOn`. This provides the simplest and clearest execution model. Step-level `dependsOn` is typically unnecessary when all operations complete synchronously and execution follows a linear path.

**Use Case: Async Coordination**

When a step must wait for an asynchronous operation to complete before proceeding, `dependsOn` establishes a join point for in-flight async work. For example, a step that requests an order status SHOULD declare `dependsOn` on the step that receives order creation status from an async order placement, even if no explicit output reference exists. This ensures the async operation completes before the dependent step executes.

**Authoring Guidance**

For async workflows, authors SHOULD use `dependsOn` to explicitly declare when a step must wait for async work to complete, regardless of output references. This is the intended use case for step-level dependencies.

**Tool Behavior**

Tools MUST respect all declared `dependsOn` relationships. Tools MUST also treat runtime expression output references (e.g., `$steps.stepId.outputs.field`) as implicit dependencies and ensure the referenced step completes before the referencing step executes.

Tools supporting only sequential execution MUST execute steps in an order that satisfies both explicit (`dependsOn`) and implicit (output reference) dependencies.

**Validation Recommendations**

Implementations SHOULD validate the following scenario:

When no `dependsOn` is used in a workflow (sequential execution model), implementations SHOULD produce an error if a step references another step's outputs where the referenced step appears later in the steps array. This indicates a forward reference that cannot be satisfied in sequential execution.
```

In the interest of continuous improvement and moving the spec forward, if a section like this, and also like the one from the previous comment to clear up step success for async steps are added to Arazzo, then I am more that happy to move forward with including this PR into the v1.1-dev branch and preparing for the 1.1 release.

nashjain: The PR is updated.

frankkilcommins (approved):
Great work @nashjain, and thanks for the additional clarification text around step `dependsOn`, and determining success for async steps.

ndenny (approved):

Issue #366, opened 2025-09-05
Proposal: Add support for JSONPath where JSON Pointer is supported


char0n:
Specifically in situations documented at https://spec.openapis.org/arazzo/v1.0.1.html#examples table.

<img width="696" height="693" alt="Image" src="https://github.com/user-attachments/assets/2fc358f3-ff20-4ae1-9923-9cc0f876346f" />

By JSONPath, I specifically refer to [RFC 9535](https://datatracker.ietf.org/doc/rfc9535/). There is no semantic overlap between JSONPath and JSON Pointer so adding a support for JSONPath WILL NOT create ambiguity as we can clearly distinguish the grammars for both of them.

Here is a concrete example where having JSONPath support would allow describing workflows that we cannot currently describe.

Let's say we have a step called `search-businesses`. This step exposes the matched business on `$steps.search-businesses.outputs.businesses`. 

Now I want to have another step called `get-businesses-details` which accepts list of business ids. If we would supported JSONPaths, we could do the following:

```
{
  "name": "business_ids",
  "in": "query",
  "value": "$steps.search-businesses.outputs.businesses#$[*].id"
}
```

With current capabilities it's not possible to chain steps like these, unless we introduce a surrogate transformation step into the workflow.

---

I can challenge this proposal and issue a PR to the spec. We have two options how to proceed forward.

### Backward compatible

We can introduce new separator for JSONPath:

`$steps.search-businesses.outputs.businesses@$[*].id` - this clearly says that `@` is followed by JSONPath and existing implementations will ignore it as they will not recognize `@` after the runtime expression.

### Backward incompatible

`$steps.search-businesses.outputs.businesses#$[*].id` - JSONPath uses the same separator (`#`) as JSON Pointer and because of no semantic overlap between JSONPath and JSON Pointer exists, implementation could clearly distinguish the intention. It will break the implementations that only expects JSON Pointer after `#` delimiter. Major release of the spec might be warranted.

char0n:
Hi @kevinduffey,

Thank for your reaction.

>  that means every tool out there will need to support BOTH or they could run in to situations where a document uses one that they dont support

Yes, that's the case.  Arazzo already requires implementations to evaluate JSONPath expressions during `Criterion Object` evaluation. And requires to evaluate JSON Pointer when plucking parts of step outputs (among others). So we're already in that situation - every compliant implementation needs to support both.

[...]

PR #374, opened 2025-10-14
fix: fix JSONPath specification ambiguity

char0n:
I assume the intention was to use RFC9535 everywhere (not the obsoleted draft)?

### DIFF
--- a/src/arazzo.md
+++ b/src/arazzo.md
@@ -706,7 +706,7 @@ Defining this object gives the ability to utilize tooling compatible with older 
 
 | Field Name | Type | Description |
 | --- | :---: | --- |
-| <a name="criterionExpressionType"></a>type | `string` | **REQUIRED**. The type of condition to be applied. The options allowed are `jsonpath` or `xpath`.                                                                                                                             |
+| <a name="criterionExpressionType"></a>type | `string` | **REQUIRED**. The type of condition to be applied. The options allowed are `jsonpath` or `xpath`. |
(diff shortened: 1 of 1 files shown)


frankkilcommins:
Hi @char0n,

The **Criterion Expression Type Object** exists to cater for situations where an author wants to intentionally deviate from the default versions of JSONPath or XPath for Criterion object expressions. In the case of JSONPath, an author would only specific this object when they specifically wanted to fallback to `draft-goessner-dispatch-jsonpath-00` instead of the default `RFC9535`.

As stated in the specification (regarding the Criterion Expression Type Object):
> An object used to describe the type and version of an expression used within a [Criterion Object](https://spec.openapis.org/arazzo/v1.0.1.html#criterion-object). If this object is not defined, then the following defaults apply:
>
>- JSONPath as described by [[RFC9535](https://spec.openapis.org/arazzo/v1.0.1.html#bib-rfc9535)]
>- XPath as described by [XML Path Language 3.1](https://www.w3.org/TR/xpath-31)
>
>Defining this object gives the ability to utilize tooling compatible with older versions of either JSONPath or XPath.

char0n:
Closing this, as this will be part of bigger framing in context of https://github.com/OAI/Arazzo-Specification/issues/366#issuecomment-3411596952 where appropriate language will have to be chosen as what selector spec versions the Arazzo implementation MUST support and what MAY be supported.

PR #409, opened 2025-11-27
feat(spec): Add support for JSONPath and XPath where JSONPointer is supported

frankkilcommins:
fixes: #366

Spec changes:

#### Expression Type Object

This replaces the previously named `Criterion Expression Type Object` and will be reused for both:

 - Criterion Object
 - Selector Object

Allowed type/version values remain:
| Type          | Allowed Versions                                 | Default   |
| ------------- | ------------------------------------------------ | --------- |
| `jsonpath`    | `rfc9535`, `draft-goessner-dispatch-jsonpath-00` | `rfc9535` |
| `xpath`       | `xpath-31`, `xpath-30`, `xpath-20`, `xpath-10`                       | `xpath-31`     |
| `jsonpointer` | `rfc6901` Just adding for completeness           | `rfc6901` |

#### Selector Object

Introduce a `Selector Object`, that can be used wherever more complex expressions than what's covered by the Arazzo ABNF grammer against structured data are required.

```
expression: string   # A valid Arazzo expression (e.g., $response.body)
selector: string     # A selector expression (e.g., $.items[0].id, /Envelope/Item)
type: string | Expression Type Object
```

Examples of usage:

```
outputs:
  userEmail:
    expression: $response.body
    selector: $.user.profile.email
    type: jsonpath
```

RequestBody example:

```
requestBody:
  contentType: application/json
  payload:
    invoiceId:
      expression: $steps.fetchXml.outputs.invoiceXml
      selector: /Invoice/Header/InvoiceNumber
      type:
        type: xpath
        version: 3.1

```

Places supporting `{expression}` for value setting have been extended to support the Selector Object.

## Schema Updates

JSON Schema updated in line with spec changes

### DIFF
--- a/src/arazzo.md
+++ b/src/arazzo.md
@@ -416,7 +416,7 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 - name: customerId
   in: query
   value:
-    expression: $inputs.customer
+    context: $inputs.customer
     selector: $.details[0].id
     type: jsonpath
 ```
@@ -910,7 +910,7 @@ A JSONPath example using an Expression Type Object:
   target: $.items[?(@.sku=='ABC123')].quantity
   targetSelectorType: jsonpath
   value:
-    expression: $steps.getInventory.outputs.payload
+    context: $steps.getInventory.outputs.payload
     selector: $.newQuantity
     type: jsonpath
 ```
@@ -923,7 +923,7 @@ An XPath example using older XPATH 3.0:
     type: xpath
     version: xpath-30
   value:
-    expression: $steps.fetchCustomerData.outputs.xml
+    context: $steps.fetchCustomerData.outputs.xml
     selector: /CustomerInfo/Id
     type:
       type: xpath


src/schemas/validation/schema.yaml
```diff
@@ -202,11 +206,15 @@ $defs:
             - $ref: '#/$defs/failure-action-object'
             - $ref: '#/$defs/reusable-object'
       outputs:
-        description: A map between a friendly name and a dynamic output value defined using a runtime expression
+        description: >
+          A map between a friendly name and a dynamic output value defined using a
+          Runtime Expression or Selector Object. Keys must match the regex: ^[a-zA-Z0-9\.\-_]+$.
```
kevinduffey: I am not a Regex guru.. but there is a '.' at the END of this that is NOT in the expression below 
patternProperties:
          '^[a-zA-Z0-9\.\-_]+$':
            type: string
            oneOf:

I asked AI (cause.. why not) if ' ^[a-zA-Z0-9\.\-_]+$.' was different than ' ^[a-zA-Z0-9\.\-_]+$' and it said yes. So I will take it's word and suggest we REMOVE the '.' at the end of the description here and elsewhere (there was another one above).

frankkilcommins: the '.' was actually the sentence punctuation. I'll remove to avoid confusion

kevinduffey: Yah I figured so (should have said that). Just thought if someone copies it there they'll pick that up and see regex may not work. It's an odd situation because a '.' has meaning in the regex expression

tests/schema/pass/parameter-with-selector.arazzo.json
```diff
@@ -0,0 +1,36 @@
...
+      "steps": [
+        {
+          "stepId": "stepA",
+          "operationId": "op1",
+          "parameters": [
+            {
+              "name": "userId",
+              "in": "query",
+              "value": {
+                "expression": "$inputs.user",
```
DmitryAnansky: This might be a bit confusing for users. There will be two different ways to describe similar things.

There is already the `context`, but this PR also suggests `expression`. Maybe I misunderstood, but it seems to perform a similar role and set the context for the condition.
And `selector` is similar to `condition`, maybe without expressions.
```
              "context": "$response.body",
              "condition": "/Envelope/Status[text()='OK']",
              "type": {
                "type": "xpath",
                "version": "xpath-31"
              }
 ```
 
Were you considering reusing already described `Criterion Object` as `Selector Object`?

Comment (frankkilcommins) [reply tests/schema/pass/parameter-with-selector.arazzo.json]: Hi @DmitryAnansky, 

Apologies in delayed response. Yes there is similarity between `Criterion Object` and `Selector Object`. The intent is to have clean separation between criteria conditional checks and traversal, but I do also see ways that we could combine. The current `condition` property in the Criterion Object does not fit well with returning something other than boolean but we could extend the object with a `selector` property which would then return the appropriate value.

There could be some benefits in this as we then could combine with power of _operators_.

Comment (frankkilcommins) [reply tests/schema/pass/parameter-with-selector.arazzo.json]: @kevinduffey @ndenny any preference here?

I have a slight preference to keep `Selector Object` and `Criterion Object` separate for 1.1, and to document their distinction clearly to avoid semantic confusion. `Selector Object` is used for extracting values (non-boolean), while `Criterion Object` is for evaluating assertions (boolean). Although they may use similar expression languages, they serve different purposes, and expanding the scope of `Criterion Object` in 1.1 could introduce additional confusion. We can revisit potential unification (such as having a formal Expression Object) in a future 2.0, where breaking changes can be introduced deliberately to support a cleaner, consolidated structure.

Comment (frankkilcommins) [reply tests/schema/pass/parameter-with-selector.arazzo.json]: @DmitryAnansky @kevinduffey @ndenny as mentioned above I have a slight preference to keep the `Selector Object` and `Criterion Object` for 1.1, but to improve consistency we could use `context` rather than `expression` for setting the root node.

File: src/schemas/validation/schema.yaml [hunk 3]
```diff
@@ -473,36 +496,79 @@ $defs:
...
+    properties:
+      expression:
+        description: A valid Runtime Expressions which MUST evaluate to structured data (e.g., `$response.body`)
+        type: string
+      selector:
+        description: A selector expression (e.g., $.items[0].id, /Envelope/Item) in the form of JSONPath expression, XPath expression, or JSON Pointer expression
+        type: string
+      type:
+        description: The selector expression type to use (e.g., `jsonpath`, `xpath`, or `jsonpointer`) or an Expression Type Object for older version support
+        oneOf:
```
nashjain: this oneOf is duplicated on line 525 as well. Would it make sense to extract it?

tests/schema/pass/parameter-with-selector.arazzo.json
kevinduffey: I agree with the use of context @frankkilcommins.

ndenny: Yeah, keeping them as separate objects makes sense for now I think. It makes things slightly easier to understand - in my own product, we have a similar bit of functionality and in the backend they're both the same thing, but in the UI, they're displayed separately.
Changing Selector to use `context` instead of `expression` makes sense - it keeps the naming consistent between the two similar cases. And then all Arazzo authors need to understand is that a selector grabs a value or object, and a condition returns a boolean.

ndenny (changes requested):

src/schemas/validation/schema.yaml
frankkilcommins: @nashjain it is extracted now into its own def

src/arazzo.md
```diff
@@ -410,6 +412,15 @@ This object MAY be extended with [Specification Extensions](#specification-exten
   value: $inputs.x-api-key
 ```
 
+```yaml
+- name: customerId
+  in: query
+  value:
+    expression: $inputs.customer
```
DmitryAnansky: Should it be `context` for `Selector Object`?

frankkilcommins: @DmitryAnansky thank you! Nice spot.

ndenny (approved):

Issue #438, opened 2026-02-19
Proposal:  add support for `querystring` Parameter Format


DmitryAnansky:
OpenAPI 3.2.0 introduced a new parameter location, `querystring`, described in  
[Extending Support for Querystring Formats](https://spec.openapis.org/oas/v3.2.0.html#extending-support-for-querystring-formats),  
along with several additional constraints:

- **`query`** — Parameters appended to the URL. For example, in `/items?id=###`, the `id` parameter is a query parameter.  
  It **MUST NOT** appear in the same operation (or in the operation’s path item) as an `in: "querystring"` parameter.

- **`querystring`** — A parameter that treats the entire URL query string as a single value. It **MUST** be specified using the `content` field, most commonly with the media type `application/x-www-form-urlencoded`, using [Encoding Objects](https://spec.openapis.org/oas/v3.2.0.html#encoding-object) in the same way as request bodies of that media type.  
  It **MUST NOT** appear more than once, and **MUST NOT** appear in the same operation (or in the operation’s path item) as any `in: "query"` parameters.

Arazzo could consider supporting this parameter directly, or alternatively document a recommended approach, so that tooling can consistently determine how to handle this syntax.

PR #443, opened 2026-03-04
feat(spec): Add querystring support compatible with OAS 3.2

frankkilcommins:
fixes: #438

### DIFF
--- a/src/arazzo.md
+++ b/src/arazzo.md
@@ -381,11 +381,11 @@ steps:
 
 #### Parameter Object
 
-Describes a single step parameter. A unique parameter is defined by the combination of a `name` and `in` fields. There are five possible locations specified by the `in` field:
+Describes a single step parameter. A unique parameter is defined by the combination of a `name` and `in` fields. There are several possible locations specified by the `in` field:
 
 - path - Used together with OpenAPI style [Path Templating](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-templating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
 - query - Parameters that are appended to the URL as individual key-value pairs. For example, in `/items?id=###`, the query parameter is `id`.
-- querystring - A parameter that treats the entire URL query string as a single value. This parameter location was introduced in [OpenAPI 3.2.0](https://spec.openapis.org/oas/v3.2.0.html) to support scenarios where the complete query string must be passed as a pre-formatted string rather than individual parameters (OpenAPI operations typically specify `application/x-www-form-urlencoded`). When a step references an operation that defines a `querystring` parameter, the value supplied MUST be a properly formatted query string (e.g., `"filter=active&sort=desc&limit=50"`). The `querystring` location cannot coexist with `query` parameters in the same operation per OpenAPI 3.2.0 constraints.
+- querystring - A parameter that treats the entire URL query string as a single value. This parameter location was introduced in [OpenAPI 3.2.0](https://spec.openapis.org/oas/v3.2.0.html) to support scenarios where the complete query string must be passed as a pre-formatted string rather than individual parameters. When a step references an operation that defines a querystring parameter, the value MUST match the media type format as expressed by the parameter's `content` field (e.g., `application/x-www-form-urlencoded`). The `querystring` location cannot coexist with `query` parameters in the same operation per OpenAPI constraints.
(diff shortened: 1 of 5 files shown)


ndenny (dismissed):

tests/schema/fail/invalid-querystring-without-value.arazzo.yaml
```diff
@@ -0,0 +1,16 @@
...
+    url: https://api.example.com/openapi.yaml
+    type: openapi
+workflows:
+  - workflowId: testInvalid
+    steps:
+      - stepId: search
+        operationId: $sourceDescriptions.api.searchItems
+        parameters:
+          - name: q
+            in: querystring
```
DmitryAnansky, line 16: It might be helpful to include another fail example that demonstrates the case where `querystring` and `query` parameters cannot coexist.

frankkilcommins: @DmitryAnansky I thought about that for quite a while before submitting the PR, and ultimately landed at the conclusion of that constraint check being an OpenAPI responsibility rather than Arazzo's.

If an OpenAPI operation described `query` and `querystring` parameters, then that would be an invalid OpenAPI Description (or Document) based on the following constraint described in OpenAPI 3.2's Parameter Object description of  parameters with `in: querystring`:

> MUST NOT appear in the same operation (or in the operation’s path-item) as any in: "query" parameters

kevinduffey (dismissed):
LGTM

src/arazzo.md
```diff
@@ -381,10 +381,11 @@ steps:
 
 #### Parameter Object
 
-Describes a single step parameter. A unique parameter is defined by the combination of a `name` and `in` fields. There are four possible locations specified by the `in` field:
+Describes a single step parameter. A unique parameter is defined by the combination of a `name` and `in` fields. There are five possible locations specified by the `in` field:
```
handrews: It might be better to just say "several possible locations" — we've started doing that in the OAS (thanks to Lorna) to avoid having these counts thrown off by future changes.

src/arazzo.md
```diff
@@ -381,10 +381,11 @@ steps:
...
-- query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`.
+- query - Parameters that are appended to the URL as individual key-value pairs. For example, in `/items?id=###`, the query parameter is `id`.
+- querystring - A parameter that treats the entire URL query string as a single value. This parameter location was introduced in [OpenAPI 3.2.0](https://spec.openapis.org/oas/v3.2.0.html) to support scenarios where the complete query string must be passed as a pre-formatted string rather than individual parameters (OpenAPI operations typically specify `application/x-www-form-urlencoded`). When a step references an operation that defines a `querystring` parameter, the value supplied MUST be a properly formatted query string (e.g., `"filter=active&sort=desc&limit=50"`). The `querystring` location cannot coexist with `query` parameters in the same operation per OpenAPI 3.2.0 constraints.
```
handrews: The phrasing here implies that only strings that look like `filter=active&sort=desc&limit=50` are "properly formatted", but that is not the case.  That is just the `application/x-www-form-urlencoded` format.  A JSON document (`application/json`) could also be valid, for example, although much of the punctuation would end up URL-encoded.

Note that using a single `in: query, style: form, explode: true, schema: {type: object}` produces a similar effect to `in: query, content: {application/x-www-form-urlencoded: ...}}`, so accessing individually named parameters would look similar.

src/arazzo.md
```diff
@@ -393,34 +394,35 @@ Describes a single step parameter. A unique parameter is defined by the combinat
...
 - name: username
   in: query
   value: $inputs.username
-```
 
-```yaml
+# Querystring Example
+- name: searchParams
+  in: querystring
+  value: "filter=active&sort=desc&limit=50" # URL-encoded format
```
handrews: If full support for `in: querystring` is intended, then an example of another format would be good.

src/arazzo.md
frankkilcommins: Good call. It caught me out previously, so no point being burned again.

src/arazzo.md
frankkilcommins: Updated the description to take above feedback into consideration and refer expectation that it should be in the format defined by the parameter's content field.

src/arazzo.md
frankkilcommins: Good shout - added an `application/json` example

ndenny (approved):

PR #451, opened 2026-03-22
feat(spec): Add $self for identity-based referencing

### LINKED ISSUE #434: Feature: Add $self for identity-based referencing
OAS 3.2 added the `$self` feature which sets the base URI and is used as the preferred `$ref` target for external referencing, which would include references from Arazzo.  To have unity across the ecosystem and ensure that an Arazzo document can be packaged for location-independent use with OADs, Arazzo should add support for a 3.2-like `$self` field and adopt similar URI / identity-based-referencing language as OAS 3.1 and 3.2.

frankkilcommins:
fixes #434


src/arazzo.md
```diff
@@ -92,10 +94,67 @@ The formats defined are:
...
-### Relative References in URLs
+### Parsing Documents
 
-Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Unless specified otherwise, relative references are resolved using the URL of the referring document.
+Each document in an Arazzo Description MUST be fully parsed in order to locate possible reference targets before attempting to resolve references.
+This includes the parsing requirements of [JSON Schema Specification Draft 2020-12](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-9), with appropriate modifications regarding base URIs as specified in [Relative References In URIs](#relative-references-in-arazzo-description-uris).
+Reference targets include the Arazzo Object's [`$self`](#arazzoSelf) field (when present).
+
+Implementations MUST NOT treat a reference as unresolvable before completely parsing all documents provided to the implementation as possible parts of the Arazzo Description.
```
handrews, line 109: There's something analogous to say here about the `url` field in the Source Description Object, which (for OAS 3.2 and later) MUST examine provided OADs form a `$self` matching that `url` field.   You arlready note further down that this field, if relative, MUST be resolved against the Arazzo document's base URI (`$self` if present) as a URI, so this is just the other end of it: once it's non-relative, you need to resolve it to a document by the target document's identity if that is different from its location.

src/arazzo.md
```diff
@@ -92,10 +94,67 @@ The formats defined are:
...
+
+If only the referenced part of a document is parsed when resolving a reference, implementations may miss the `$self` field or Source Description URIs, causing references to resolve to unintended locations. The resulting behavior of fragmentary parsing is _undefined_ and NOT RECOMMENDED.
```
handrews: This language is present in the OAS because in Swagger/OAS 2.0, the expectation _was_ to parse targets in isolation (more or less... it's not quite spelled out but that's how it often worked).  And then it was completely ambiguous in 3.0.  So 3.1.1+ and 3.2 have this language to let people know something changed.  If Arazzo documents were always parsed as whole documents, this is probably more confusing than helpful.  I have no idea what Arazzo tools do today, though.

src/arazzo.md
```diff
@@ -92,10 +94,67 @@ The formats defined are:
...
+
+#### Establishing the Base URI
+
+Relative URI references are resolved using the appropriate base URI, which MUST be determined in accordance with [RFC3986 Section 5.1.1 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.1).
+
+The base URI for resolving relative references within an Arazzo Description is determined as follows:
+
+- If the [`$self`](#arazzoSelf) field is present and is an absolute URI, the base URI is the `$self` URI.
+- If the [`$self`](#arazzoSelf) field is present and is a relative URI-reference, the base URI is the result of resolving `$self` against the retrieval URI (or other applicable base URI per RFC3986 Section 5.1).
+- If the [`$self`](#arazzoSelf) field is not present, the base URI is the retrieval URI of the Arazzo Description document.
```
handrews: This wording makes it sound like the retrieval URI is the next place to look, which (per RFC3986) it is not.  I'm nitpicking on this because it's a very persistent misconception, which is why the OAS says:

> If $self is a relative URI reference, it is resolved against the next possible base URI source ([[RFC3986](https://spec.openapis.org/oas/v3.2.0.html#bib-rfc3986)] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2)) before being used for the resolution of other relative URI references.

and then talks about the retrieval URI as the most common (but not next) place to _find_ the base URI.

If you want use cases for the non-`$self`, non-retrieval URIs I can provide them.

src/arazzo.md
```diff
@@ -92,10 +94,67 @@ The formats defined are:
...
+
+For examples demonstrating base URI determination and reference resolution, see [Appendix B: Examples of Base URI Determination and Reference Resolution](#appendix-b-examples-of-base-uri-determination-and-reference-resolution).
+
+#### Resolving URI Fragments
+
+If a URI contains a fragment identifier, then the fragment MUST be resolved per the fragment resolution mechanism of the referenced document.
+
+For JSON or YAML documents (including OpenAPI or AsyncAPI descriptions referenced via Source Descriptions), the fragment identifier SHOULD be interpreted as a JSON Pointer as per [RFC6901](https://tools.ietf.org/html/rfc6901).
```
handrews: I see this came from the OAS, but I just realized we have a slight bug here:  Schema Objects can also declare plain name fragments.  I'm not sure it's worth worrying about right now as I am not sure off the top of my head how to explain that part, but perhaps an issue to fix it before shipping?  I just need to think on it for a little bit.

src/arazzo.md
```diff
@@ -22,7 +22,8 @@ The Arazzo Specification can articulate these workflows in a human-readable and
   - [Format](#format)
   - [Arazzo Description Structure](#arazzo-description-structure)
   - [Data Types](#data-types)
-  - [Relative References in URLs](#relative-references-in-urls)
+  - [Parsing Documents](#parsing-documents)  
+  - [Relative References in Arazzo Description URIs](#relative-references-in-arazzo-description-uris)
   - [Schema](#schema)
```
handrews, line 33: I think you accidentally dropped "Relative References in API URLs" (that section looks good, btw, nice distinctions regarding Arazzo docs vs OAD docs vs APIs vs runtime expressions).

src/arazzo.md
```diff
@@ -1064,3 +1125,43 @@ The proposed MIME media type for Arazzo documents (e.g. workflows) that require
...
+  - name: petstore
+    url: ../specs/petstore.yaml  # Resolves to https://api.example.com/specs/petstore.yaml
+    type: openapi
+```
+
+The relative URL `../specs/petstore.yaml` resolves against the `$self` base URI (`https://api.example.com/workflows/`), producing `https://api.example.com/specs/petstore.yaml`, regardless of the retrieval URI.
+
+### Base URI From the Retrieval URI (No `$self`)
+
+If the same document does not define `$self`:
```
handrews, line 1183: I think the other examples from [OAS Appendix F](https://spec.openapis.org/oas/v3.2.0.html#appendix-f-examples-of-base-uri-determination-and-reference-resolution) are needed here, as the retrieval URI is not the only possible additional source.

Adding more sources shouldn't break compatibility, because in the exceedingly unlikely scenario that anyone was actually parsing something where there's an encapsulating entity (e.g. `multipart/related` hypermedia archive format), there's no obvious "retrieval URI" anyway, so they're doing something else not defined by Arazzo.  I think the bits about handling `$id` in Schema Objects are relevant for you as well.

handrews (changes requested):
Thanks for doing this!  Most of it looks great, a few minor bits plus RFC3986 base URI determination being a bit more complex than is stated here.  Plus making it explicit that referencing an OAD by that OAD's `$self` requires examining potential target OADs for `$self`.

src/arazzo.md
frankkilcommins: Added new "Identity-Based Referencing" subsection under "Parsing Documents" that explains implementations must examine target documents for matching $self values when resolving references.

src/arazzo.md
frankkilcommins: Updated "Establishing the Base URI" section to properly reference RFC3986 sections 5.1.1-5.1.4 and use "next possible base URI source" language instead of oversimplifying.

src/arazzo.md
frankkilcommins: Clarified that Arazzo uses JSON Pointer fragments, and reference resolution within referenced OpenAPI/AsyncAPI documents must follow those specs' rules.

src/arazzo.md
frankkilcommins: added missing subsections to TOC

src/arazzo.md
frankkilcommins: Added four new sections to Appendix B covering encapsulating entities, application defaults, resolving relative $self, and identity vs location examples.

src/arazzo.md
handrews: @frankkilcommins Arazzo uses JSON Schema draft 2020-12 in its Schema Objects, correct?  If so, any reference into an Arazzo Schema Object needs to be able to handle plain name fragments as well.

src/arazzo.md
handrews: I almost wonder if we should just have a separate document explaining all of this for all of our specifications... a thought for another time :-)

src/arazzo.md
frankkilcommins: @handrews to clarify, I should probably enhance with the following:

>  For JSON Schema objects used in workflow `inputs` and `components.inputs`, references via `$ref` MUST support both JSON Pointer fragments and plain-name fragments (defined by `$anchor`) as specified by [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-01.html).

handrews: @frankkilcommins actually I just looked at OAS 3.2 and we say _almost_ the same thing you're saying here, and just kind-of hand-wave the plain-name fragment things under the "should be resolved per the fragment resolution mechanism of the referenced document" clause (that Arazzo says "MUST" instead of "should" here is an improvement!).

I think the difference that is critical is that your current wording implies that OpenAPI descriptions always use JSON Pointers.  So you could either be more clear about Schema Objects (in which case what you have proposed seems fine) or you could be more vague about all of it.

frankkilcommins: @handrews I had just committed above when this comment came through ;-)

src/arazzo.md
```diff
@@ -1064,3 +1155,127 @@ The proposed MIME media type for Arazzo documents (e.g. workflows) that require
...
+  title: Pet Purchase Workflow
+  version: 1.0.0
+sourceDescriptions:
+  - name: petstore
+    url: ../specs/petstore.yaml
+```
+
+Retrieved from `https://api.example.com/v2/api-description.yaml`:
+
+- First, resolve `$self` relative reference: `workflows/purchase.arazzo.yaml` against base `https://api.example.com/v2/`, which then resolves to `https://api.example.com/v2/workflows/purchase.arazzo.yaml`
```
DmitryAnansky: Could you please clarify how we are reading this base `https://api.example.com/v2/`?
Should it be set externally?

handrews: @DmitryAnansky good catch!  There are two parts of my answer here:  One is that in this example the base URI is the retrieval URI, and it is implementation-specific how that is determined (but in practice, yes, tools should allow setting it externally since you often don't want tools making random network requests for security reasons).  

For the other part...

@frankkilcommins the base URI here should be `https://api.example.com/v2/api-description.yaml`.  The first step of _using_ that as a base URI with a path-relative URI reference like `workflows/purchase.arazzo.yaml` is to chop off the trailing path segment (`api-description.yaml`), but that's part of the resolution _process_.  `https://api.example.com/v2/api-description.yaml` is the base URI.

src/arazzo.md
handrews: Shouldn't it be at the same level as "Relative References in Arazzo Description URIs"?  It's not really a sub-topic of that.

src/arazzo.md
```diff
@@ -1064,3 +1155,127 @@ The proposed MIME media type for Arazzo documents (e.g. workflows) that require
...
+  ]
+}
+--example--
+```
+
+The `Content-Location` header provides the base URI (`https://api.example.com/workflows/purchase.arazzo.json`), so `../specs/petstore.json` resolves to `https://api.example.com/specs/petstore.json` even without a `$self` field.
+
+### Application-Specific Default Base URI
+
+Per [RFC3986 Section 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.4), applications may define default base URIs. For example, a workflow orchestration platform might establish `https://workflows.example.com/` as the default base for all Arazzo Descriptions loaded without explicit retrieval URIs.
```
handrews: While I _think_ this is technically in compliance with RFC3986, I usually see this handled as a per-document thing, often an auto-generated URN, or a URL with a fixed base (such as `https://workflows.example.com/` and then a randomized final component.

I'm not making this an objection, it just feels a bit odd in the context of `$self` which is document-specific.  `https://workflows.example.com/` might feel like a "directory", but really in URI terms there is a final path segment of an empty string after that `/`.  It's not conceptually different from `https://workflows.example.com/123456`, it's just that we, as humans, think of the empty string as not being there.  This is also why I keep commenting that that trailing path component is part of the base URI.. when you truncate it to a trailing `/` you're not _really_ truncating, you're replacing a non-empty string with an empty string.  (and yes, this is a ridiculously pedantic point, but in theory it could matter... don't ask me for an example off the top of my head though)

Meh... make of this what you will.  The use case you're implying here is certainly valid, and more helpful than a random URN.

src/arazzo.md
```diff
@@ -1064,3 +1155,127 @@ The proposed MIME media type for Arazzo documents (e.g. workflows) that require
...
+  title: Pet Purchase Workflow
+  version: 1.0.0
+sourceDescriptions:
+  - name: petstore
+    url: ../specs/petstore.yaml  # Resolves to https://api.example.com/specs/petstore.yaml
+    type: openapi
+```
+
+The relative URL `../specs/petstore.yaml` resolves against the `$self` base URI (`https://api.example.com/workflows/`), producing `https://api.example.com/specs/petstore.yaml`, regardless of the retrieval URI.
+
```
handrews: As noted in another thread, the base URI includes the final path component, and is `https://api.example.com/workflows/purchase.arazzo.yaml`.  Trimming off the `purchase.arazzo.yaml` is part of the resolution algorithm.

src/arazzo.md
```diff
@@ -1064,3 +1155,127 @@ The proposed MIME media type for Arazzo documents (e.g. workflows) that require
...
+  version: 1.0.0
+```
+
+This document might be:
+
+- Retrieved from `https://cdn.example.com/cache/abc123.yaml`
+- Retrieved from `file:///local/dev/purchase.arazzo.yaml`
+- Embedded in a `multipart/related` response
+
+In all cases, references to this Arazzo Description MUST use `https://workflows.example.com/canonical/purchase.arazzo.yaml` (the `$self` value), not the retrieval location. This ensures that references remain stable even when the document is mirrored, cached, or moved.
```
handrews, line 1285: This is a nice section to give a bit of the "why" here.

If you want to go a bit further, you might note that deploying a set of documents behind a firewall (or on an air-gapped network) is a reason to scan for `$self` values.  That way you can locate documents from a provided set without having to make a network call that your firewall or other configuration prevents.

Feel free to take this suggestion or not, it's not at all a blocker and is fine to leave out.

handrews (changes requested):
Just a few more minor things... thanks for all of the work on this!

src/arazzo.md
frankkilcommins: You're right and it's better to clear rather than just assuming how it will be interpreted. I've added more explicit language.

src/arazzo.md
frankkilcommins: I added a sentence or two on the topic:

> Identity-based referencing via `$self` is particularly valuable in security-restricted environments. When deploying document sets behind firewalls or on air-gapped networks, implementations can scan for `$self` values to locate documents from a provided collection without making network requests that security policies might prevent. This enables reference resolution in environments where external network access is restricted or prohibited.

src/arazzo.md
handrews: Thanks!  The new version looks good and keeps the best / most realistic aspects of the old version.

handrews (dismissed):

src/arazzo.md
```diff
@@ -110,6 +199,7 @@ This is the root object of the [Arazzo Description](#arazzo-description).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="arazzoVersion"></a>arazzo | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the Arazzo Specification that the Arazzo Description uses. The `arazzo` field MUST be used by tooling to interpret the Arazzo Description. |
+| <a name="arazzoSelf"></a>$self | `string` | A URI-reference for the Arazzo Description. This string MUST be in the form of a URI-reference as defined by [RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1). When present, this field provides the self-assigned URI of this Arazzo Description, which also serves as its base URI in accordance with [RFC3986 Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1) for resolving relative references within this document. The `$self` URI MUST NOT contain a fragment identifier. Arazzo Description documents SHOULD include a `$self` field to ensure portable, unambiguous reference resolution. |
```
handrews: > Arazzo Description documents SHOULD include a `$self` field to ensure portable, unambiguous reference resolution.

I just noticed this.  While it is my personal preference, there are times when you really do want things to be relative to wherever you are, so I hesitate to include a SHOULD here.  A SHOULD is pretty strong.

frankkilcommins: softened the statement.

handrews (approved):
@frankkilcommins it's interested that you softened that language with a "can" while @baywet made it a "MAY" over in Overlays.  I have no idea which is better and am fine with either, I just thought I'd make a note of it.

ndenny (approved):

Issue #437, opened 2026-02-18
List or object usage in runtime expressions embedded in strings


DmitryAnansky:
## What/Why/How?

HI All,
I’ve noticed that the expected behavior for runtime expression evaluation, particularly when dealing with lists or objects, is not clearly defined.
Technically, it is possible to produce a Step output in the following form:
```yaml
    steps:
      - stepId: get-products
        operationId: listOfProducts
        outputs:
          products: $response.body#/items
```
Lets consider that response items is a list of objects:

```json
[
   {
       "name": "item1",
       "id": 0
   },
   {
       ...
   },
]
```

It is also possible to use the Step output as runtime expressions embedded in string,
like in some specification examples:

```yaml
  contentType: application/xml
  payload: |
    <order>
      <items>{$steps.get-products.outputs.products}</items>
    </order>
```

```
  contentType: application/x-www-form-urlencoded
  payload: 
    products: {$steps.get-products.outputs.products}
```

How should we interpret the expression in case when it is a `list` / `list of objects` / `object` ? 
Should it be stringified value, or should we throw an error in this case?

## Reference

Might be related to https://github.com/OAI/Arazzo-Specification/issues/424

handrews:
> Scalar values (string, number, boolean, null) are converted to their string representation.

String representation according to what rules?  This is not a trivial problem, as noted in OAS [Appendix B: Data Type Conversion](https://spec.openapis.org/oas/v3.2.0.html#appendix-b-data-type-conversion).

> If the value is a parsed structure (e.g., a JSON object or array from a parsed response, or workflow input), it MUST be serialized as JSON per RFC 8259.

I don't know if this is possible, but it would be better if it were possible to indicate the serialization format, as the OAS Encoding Object allows for `application/x-www-form-urlencoded`.

Issue #428, opened 2026-01-22
Ambiguity in $components runtime expressions


char0n:
We have ambiguity for: `"$components." name / "$components.parameters." parameter-name`

Both parse as `$components.` + `name`:

```
$components.parameters.foo → name = "parameters.foo"
$components.schemas.Pet    → name = "schemas.Pet"
```

The "$components.parameters." parameter-name alternative is never reached because "$components." name matches first with the greedy name = *( CHAR ) rule.

Issue #427, opened 2026-01-22
Ambiguity in $sourceDescription.* runtime expressions


char0n:
### Problem

Current format `$sourceDescriptions.<name>.<reference>` is ambiguous - `<reference>` could be:
 - A field from SourceDescription object (like `url`)
 - An `operationId` from OpenAPI source
 - A `workflowId` from Arazzo source

If we have `operationId="url"` in source OpenAPI Document, or `workflowId="url"` in Arazzo Document, it becomes ambiguous. 

### Proposed Format (Unambiguous)

 `$sourceDescriptions.<sourceName>.<fieldName>`                    # field access
 `$sourceDescriptions.<sourceName>.operation.<operationId>`        # operationId reference
 `$sourceDescriptions.<sourceName>.workflow.<workflowId>`          # workflowId reference (allows us to validate workflowId format)

 Examples:
 - `$sourceDescriptions.petStore.url` → field access
 - `$sourceDescriptions.petStore.operation.getPets` → operationId
 - `$sourceDescriptions.petStore.workflow.loginFlow` → workflowId

Issue #426, opened 2026-01-22
Introduce complete grammar for parsing expressions


char0n:
Hi, 

I'm proposing to introduce a complete grammars for describing syntax of following runtime expression types:

``` 
"$steps." name 
"$workflows." name 
"$sourceDescriptions." name 
"$components." name
```

These types will be replaced by

``` 
"$steps." steps-name 
"$workflows." workflows-name 
"$sourceDescriptions." source-descriptions-name 
"$components." components-name
```

Full grammar for above non-terminals is defined below:

```abnf
; Secondary grammar for parsing $steps name part
; Format: {stepId}.{field}.{subField}[#/{jsonPointer}]
steps-name       = steps-id "." steps-field "." steps-sub-field ["#" json-pointer]
steps-id         = 1*(ALPHA / DIGIT / "_" / "-")
steps-field      = "outputs"
steps-sub-field  = 1*(ALPHA / DIGIT / "." / "-" / "_")

; Secondary grammar for parsing $workflows name part
; Format: {workflowId}.{field}.{subField}[#/{jsonPointer}]
workflows-name       = workflows-id "." workflows-field "." workflows-sub-field ["#" json-pointer]
workflows-id         = 1*(ALPHA / DIGIT / "_" / "-")
workflows-field      = "inputs" / "outputs"
workflows-sub-field  = 1*(ALPHA / DIGIT / "." / "-" / "_")

; Secondary grammar for parsing $sourceDescriptions name part
; Format: {sourceName}.{reference}
; reference can be operationId (unconstrained) or workflowId (constrained)
source-descriptions-name        = source-descriptions-source-name "." source-descriptions-reference
source-descriptions-source-name = 1*(ALPHA / DIGIT / "_" / "-")
source-descriptions-reference   = 1*CHAR

; Secondary grammar for parsing $components name part
; Format: {field}.{subField}
; Allowed fields: inputs, parameters, successActions, failureActions
components-name      = components-field "." components-sub-field
components-field     = "parameters" / "successActions" / "failureActions"
components-sub-field = 1*(ALPHA / DIGIT / "." / "-" / "_")
```

PR #454, opened 2026-03-23
feat(runtime-expressions): improve ABNF grammar clarity

frankkilcommins:
- Define CHAR to exclude curly braces for unambiguous parsing (#424)
- Add expression-string and embedded-expression grammar (#425)
- Specialize reference types by context (#426)
- Add Source Description expression resolution priority (#427)
- Require explicit component types in $components (#428)
- Clarify type conversion in embedded expressions (#437)
- Fix invalid examples in spec (conditions, stepId references)
- Update examples table with nested inputs and component actions

$components now requires explicit component type (parameters/successActions/failureActions). Generic components.name pattern removed. Note: This was already semantically invalid per spec.

fixes: #424
fixes: #425
fixes: #426
fixes: #428
fixes: #437

resolves: #427

### DIFF
--- a/examples/1.0.0/bnpl-arazzo.yaml
+++ b/examples/1.0.0/bnpl-arazzo.yaml
@@ -167,10 +167,10 @@ workflows:
       contentType: application/json
       payload: |
         {
-          "firstName": "{$inputs.customer.firstName}",
-          "lastName": "{$inputs.customer.lastName}",
-          "dateOfBirth": "{$inputs.customer.dateOfBirth}",
-          "postalCode": "{$inputs.customer.postalCode}"
+          "firstName": "{$inputs.customer#/firstName}",
+          "lastName": "{$inputs.customer#/lastName}",
+          "dateOfBirth": "{$inputs.customer#/dateOfBirth}",
+          "postalCode": "{$inputs.customer#/postalCode}"
           "termsAndConditionsAccepted": true
         }
     successCriteria:
@@ -194,7 +194,7 @@ workflows:
       contentType: application/json
       payload: | 
         {
-          "enrolledCustomer": "{$inputs.customer.uri}",
+          "enrolledCustomer": "{$inputs.customer#/uri}",
           "newCustomer": "{$steps.createCustomer.outputs.customer}",
           "products": "{$steps.checkLoanCanBeProvided.outputs.eligibleProducts}"
         }
--- a/examples/1.0.0/bnpl-arazzo.yaml
+++ b/examples/1.0.0/bnpl-arazzo.yaml
@@ -170,7 +170,7 @@ workflows:
           "firstName": "{$inputs.customer#/firstName}",
           "lastName": "{$inputs.customer#/lastName}",
           "dateOfBirth": "{$inputs.customer#/dateOfBirth}",
-          "postalCode": "{$inputs.customer#/postalCode}"
+          "postalCode": "{$inputs.customer#/postalCode}",
           "termsAndConditionsAccepted": true
         }
     successCriteria:
--- a/src/arazzo.md
+++ b/src/arazzo.md
@@ -1005,7 +1005,7 @@ The runtime expression is defined by the following [ABNF](https://tools.ietf.org
   ; JSON Pointer (RFC 6901)
   json-pointer = *( "/" reference-token )
   reference-token = *( unescaped / escaped )
-  unescaped = %x00-2E / %x30-7A / %x7C / %x7E-10FFFF
+  unescaped = %x00-2E / %x30-7A / %x7C / %x7F-10FFFF
       ; Excludes / (%x2F), { (%x7B), } (%x7D), and ~ (%x7E)
   escaped = "~" ( "0" / "1" )
       ; representing '~' and '/', respectively
--- a/src/arazzo.md
+++ b/src/arazzo.md
@@ -946,8 +946,8 @@ The runtime expression is defined by the following [ABNF](https://tools.ietf.org
       "$statusCode" /
       "$request." source /
       "$response." source /
-      "$inputs." input-reference /
-      "$outputs." output-reference /
+      "$inputs." inputs-reference /
+      "$outputs." outputs-reference /
       "$steps." steps-reference /
       "$workflows." workflows-reference /
       "$sourceDescriptions." source-reference /
@@ -963,25 +963,26 @@ The runtime expression is defined by the following [ABNF](https://tools.ietf.org
   body-reference = "body" ["#" json-pointer ]
 
   ; Input/Output references
-  input-reference = input-name [ "#" json-pointer ]
-  output-reference = output-name [ "#" json-pointer ]
-  input-name = identifier
-  output-name = identifier
+  inputs-reference = input-name [ "#" json-pointer ]
+  outputs-reference = output-name [ "#" json-pointer ]
+  input-name = name
+  output-name = name
 
   ; Steps expressions
   steps-reference = step-id ".outputs." output-name [ "#" json-pointer ]
-  step-id = identifier
+  step-id = identifier-strict
 
   ; Workflows expressions
   workflows-reference = workflow-id "." workflow-field "." field-name [ "#" json-pointer ]
-  workflow-id = identifier
+  workflow-id = identifier-strict
   workflow-field = "inputs" / "outputs"
-  field-name = identifier
+  field-name = name
 
   ; Source descriptions expressions
   source-reference = source-name "." reference-id
-  source-name = identifier
-  reference-id = identifier
+  source-name = identifier-strict
+  reference-id = 1*CHAR
+      ; operationIds have no character restrictions in OpenAPI/AsyncAPI
       ; Resolution priority defined in spec text: (1) operationId/workflowId, (2) field names
 
   ; Components expressions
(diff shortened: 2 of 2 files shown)


examples/1.0.0/bnpl-arazzo.yaml
```diff
@@ -167,10 +167,10 @@ workflows:
...
       payload: |
         {
-          "firstName": "{$inputs.customer.firstName}",
-          "lastName": "{$inputs.customer.lastName}",
-          "dateOfBirth": "{$inputs.customer.dateOfBirth}",
-          "postalCode": "{$inputs.customer.postalCode}"
+          "firstName": "{$inputs.customer#/firstName}",
+          "lastName": "{$inputs.customer#/lastName}",
+          "dateOfBirth": "{$inputs.customer#/dateOfBirth}",
+          "postalCode": "{$inputs.customer#/postalCode}"
```
DmitryAnansky: It would be great to include more complex and diverse examples demonstrating the application of ABNF syntax.

src/arazzo.md
```diff
@@ -939,23 +939,91 @@ A runtime expression allows values to be defined based on information that will
...
+  reference-id = identifier
+      ; Resolution priority defined in spec text: (1) operationId/workflowId, (2) field names
+
+  ; Components expressions
+  components-reference = component-type "." component-name
+  component-type = "parameters" / "successActions" / "failureActions"
+  component-name = identifier
+
+  ; Identifier rule
+  identifier = 1*( ALPHA / DIGIT / "." / "-" / "_" )
```
char0n, line 998: The PR's identifier = 1*( ALPHA / DIGIT / "." / "-" / "_" ) is used for all IDs (stepId, workflowId, sourceDescriptionName, component keys, input/output names). But the spec defines two
  different patterns:
                                                                                                                                                                                                  
  - stepId, workflowId, sourceDescriptionName: SHOULD [A-Za-z0-9_\-]+ (no dot)                                                                                                                    
  - Components keys: MUST ^[a-zA-Z0-9\.\-_]+$ (with dot)                                                                                                                                          
                                                                                                                                                                                                  
  A single shared identifier rule conflates these — it allows dots in step/workflow IDs where the spec says they shouldn't be, and it's only SHOULD-level enforcement anyway. Separate rules would
   be more faithful to the spec's intent.

src/arazzo.md
```diff
@@ -939,23 +939,91 @@ A runtime expression allows values to be defined based on information that will
...
+  step-id = identifier
+
+  ; Workflows expressions
+  workflows-reference = workflow-id "." workflow-field "." field-name [ "#" json-pointer ]
+  workflow-id = identifier
+  workflow-field = "inputs" / "outputs"
+  field-name = identifier
+
+  ; Source descriptions expressions
+  source-reference = source-name "." reference-id
```
char0n: `source-reference` is too restrictive with `identifier`
                                                                                              
  The proposed grammar uses:
```  
  source-reference = source-name "." reference-id                                             
  reference-id = identifier                      
```
                                                                                              
  The `<reference>` part can be an operationId from an OpenAPI description or a workflowId from 
  an Arazzo document. OpenAPI does not constrain operationId to any specific character set —  
  it's just a string. This means operationIds like get/pets, get pets, or create-user@v2 are  
  technically valid in OpenAPI but would be rejected by the identifier rule.                  
                  
  I'd suggest using a less restrictive rule for reference-id — something like 1*CHAR (any     
  character except `{ and }`) — to avoid rejecting valid OpenAPI operationIds.

src/arazzo.md
```diff
@@ -939,23 +939,91 @@ A runtime expression allows values to be defined based on information that will
...
-      tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
-        "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
+  ; Top-level expression
+  expression = (
+      "$url" /
+      "$method" /
+      "$statusCode" /
+      "$request." source /
+      "$response." source /
+      "$inputs." input-reference /
```
char0n: Plural for consistency with `$steps` or `$workflows` 

```suggestion
      "$inputs." inputs-reference /
```

src/arazzo.md
```diff
@@ -939,23 +939,91 @@ A runtime expression allows values to be defined based on information that will
...
+      "$outputs." output-reference /
```
char0n:
```suggestion
      "$outputs." outputs-reference /
```

src/arazzo.md
```diff
@@ -939,23 +939,91 @@ A runtime expression allows values to be defined based on information that will
...
+      ; Alphanumeric with dots, hyphens, underscores
+      ; Matches recommended pattern [A-Za-z0-9_\-]+ from spec
+
+  ; Legacy 'name' rule (retained for query/path references)
+  name = *( CHAR )
+
+  ; JSON Pointer (RFC 6901)
+  json-pointer = *( "/" reference-token )
+  reference-token = *( unescaped / escaped )
+  unescaped = %x00-2E / %x30-7D / %x7F-10FFFF
```
char0n: unescaped in json-pointer still includes { and } — breaks embedded expression parsing
                                                                                        
  The PR correctly excludes { and } from the CHAR rule for unambiguous embedded
  expression parsing, but the unescaped rule in json-pointer still uses %x30-7D, which  
  includes } (%x7D) and { (%x7B).                           
                                                                                        
  This means an embedded expression like {$request.body#/status} or                     
  {$steps.someStepId.outputs.pets#/0/id} cannot be reliably parsed — the json-pointer's
  unescaped will consume the closing }, making it impossible to determine where the     
  expression ends.                                          

  The fix is to change unescaped from:                                                  
  unescaped = %x00-2E / %x30-7D / %x7F-10FFFF
  to:                                                                                   
  unescaped = %x00-2E / %x30-7A / %x7C / %x7E-10FFFF                                    
      ; %x2F ('/'), %x7E ('~'), %x7B ('{'), %x7D ('}') are excluded                     
                                                                                        
  This is a minor deviation from RFC 6901, but { and } in JSON Pointer reference tokens 
  are extremely rare in practice, and without this fix the expression-string grammar    
  cannot work correctly for any expression containing a json-pointer.                   
                                                                                        
  We validated this in our ABNF parser implementation at                                
  https://github.com/swaggerexpert/arazzo-runtime-expression — after making this change,
   all expressions with json-pointers work correctly in both standalone and embedded    
  contexts.

src/arazzo.md
```diff
@@ -939,23 +939,91 @@ A runtime expression allows values to be defined based on information that will
...
+  component-type = "parameters" / "successActions" / "failureActions"
+  component-name = identifier
+
+  ; Identifier rule
+  identifier = 1*( ALPHA / DIGIT / "." / "-" / "_" )
+      ; Alphanumeric with dots, hyphens, underscores
+      ; Matches recommended pattern [A-Za-z0-9_\-]+ from spec
+
+  ; Legacy 'name' rule (retained for query/path references)
+  name = *( CHAR )
```
char0n, line 1002: name isn't legacy. It's the correct rule for query and path references because query parameter names and path parameter names are user-defined and can contain any valid character.

src/arazzo.md
frankkilcommins: updated to:

```
  ; Source descriptions expressions
  source-reference = source-name "." reference-id
  source-name = identifier-strict
  reference-id = 1*CHAR
      ; operationIds have no character restrictions in OpenAPI/AsyncAPI
      ; Resolution priority defined in spec text: (1) operationId/workflowId, (2) field names
```

src/arazzo.md
```diff
@@ -963,45 +963,50 @@ The runtime expression is defined by the following [ABNF](https://tools.ietf.org
...
   name = *( CHAR )
+      ; Allows unrestricted characters for query/path parameter names and field references
 
   ; JSON Pointer (RFC 6901)
   json-pointer = *( "/" reference-token )
   reference-token = *( unescaped / escaped )
-  unescaped = %x00-2E / %x30-7D / %x7F-10FFFF
-      ; %x2F ('/') and %x7E ('~') excluded from 'unescaped'
+  unescaped = %x00-2E / %x30-7A / %x7C / %x7E-10FFFF
+      ; Excludes / (%x2F), { (%x7B), } (%x7D), and ~ (%x7E)
```
DmitryAnansky, line 1009: Will it make sense to explicitly define some other literals, like boolean, undefined, null?
They can be used in runtime expressions.

```
// Boolean literals for true and false
Boolean
  = "true" / "false"

// Null literal
Null
  = "null"

Undefined
  = "undefined"
```

char0n:
Hi @frankkilcommins,

  Hi Frank,                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                             
  Great progress on the grammar updates — the `identifier-strict` split, `reference-id = 1*CHAR`, and json-pointer `unescaped` fix all look good.                                                                                                                                            
                                                                                                                                                                                                                                                                                             
  I noticed a couple of issues with the latest changes.                                                                                                                                                                                                                                      
                            
  ### `input-name = name` breaks json-pointer parsing                                                                                                                                                                                                                                        
                            
  Since `name = *( CHAR )` and `CHAR` includes `#` (%x23, within the `%x00-7A` range), the greedy `*( CHAR )` will consume the `#` as part of the name. This means the optional `["#" json-pointer]` in rules like:                                                                          
                            
  ```abnf                                                                                                                                                                                                                                                                                    
  inputs-reference = input-name [ "#" json-pointer ]
```
                                                                                                                                                                                                                                                                                             
  would never match, because the parser consumes customer#/firstName entirely as the input-name, leaving nothing for the json-pointer.                                                                                                                                                       
                                                                                                                                                                                                                                                                                             
  For example, $inputs.customer#/firstName would parse with input-name = "customer#/firstName" and no json-pointer — which defeats the purpose of adding json-pointer support to inputs/outputs.                                                                                             
                            
  The previous version using identifier worked because identifier doesn't include #, so the parser correctly stops at # and matches the json-pointer. The same issue applies to output-name = name and field-name = name in workflows-reference.                                             
                            
  The fix is either:                                                                                                                                                                                                                                                                         
  1. Keep identifier for these rules (what I use in my implementation)
  2. Define a new rule like name-no-hash that is CHAR minus # — more permissive than identifier but still allows the json-pointer delimiter to work                                                                                                                                          
                                                                                                                                                   
  Option 1 is simpler. Option 2 is more permissive but requires defining a new character class.                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                             
 `CHAR` redefinition diverges from OpenAPI                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                             
  The PR redefines CHAR as a simple character range (%x00-7A / %x7C / %x7E-10FFFF), dropping the RFC 7159 JSON string definition with escape sequences. OpenAPI's runtime expression ABNF uses the RFC 7159-based CHAR with unescape / escape rules. Since Arazzo builds on top of OpenAPI and shares the runtime expression concept, redefining CHAR introduces a divergence — a bare `\` becomes valid in Arazzo but not in OpenAPI, and escape sequences like `\n` or `\uXXXX` are no longer recognized. I'd recommend keeping the RFC 7159-based definition for interoperability.

src/arazzo.md
frankkilcommins: The literals (boolean, null, undefined) are used in Criterion Object conditions alongside runtime expressions, not within runtime expressions themselves. These are already defined in the Criterion Object section as part of the condition syntax. The ABNF grammar here specifically defines the structure of runtime expressions (`$inputs.foo`, `$statusCode`, etc.), which reference values but don't contain literals. The condition evaluation syntax is separate from the runtime expression parsing syntax.

Do you have scenarios that you're thinking of? I'd tend to scope that under a different enhancement issue if warranted.

char0n (approved):
The ABNF is now complete and not ambiguous.

My last suggestion would be to probably rename `reference-id` to `source-reference-id` to remove ambiguity in naming.

ndenny (approved):

PR #465, opened 2026-04-15
chore: merge v1.0-dev fixes into v1.1-dev

frankkilcommins:
Consolidating v1.0.2 fixes into v1.1.0.

Our plan is to just proceed with a `v1.1.0` release rather than also releasing a `v1.0.2` version.


frankkilcommins:
Replaced by #466 due to branch protection policies that I don't want to by-pass

PR #476, opened 2026-04-29
Add missing schema validation fixtures for async, OpenAPI, and workflow steps

nashjain:
## Summary

Adds fixture coverage for schema validation branches that were already enforced by the Arazzo v1.1 schema but were not explicitly exercised by the test suite.

This PR expands coverage in four areas:

- Async step validation
- OpenAPI step validation
- Workflow step validation
- Characterization of current schema-permitted behavior

### Async fixtures
Adds coverage for:

- `operationId` and `channelPath` mutual exclusivity
- required `action` on async `channelPath` steps
- `dependsOn` empty array rejection
- invalid `dependsOn` reference syntax rejection
- duplicate `dependsOn` entry rejection
- valid `receive` step without `correlationId`
- reusable parameter references on async steps

### OpenAPI fixtures
Adds coverage for:

- valid `operationPath`-only steps
- reusable parameter references on OpenAPI steps
- invalid steps with both `operationId` and `operationPath`
- invalid steps missing both `operationId` and `operationPath`
- duplicate parameter entry rejection

### Workflow step fixtures
Adds coverage for:

- minimal valid workflow steps
- reusable parameter references on workflow steps
- invalid workflow steps missing `workflowId`
- duplicate workflow-step parameter rejection
- rejection of async-only fields like `action` on workflow steps
- rejection of `channelPath` on workflow steps

### Characterization fixtures
Adds pass fixtures that document current validator behavior where structural schema validation succeeds even though deeper semantic validation would still be required:

- `asyncapi` operation step without `action`
- OpenAPI operation step with async `action`
- OpenAPI receive step with `correlationId`
- permissive `operationPath` string handling
- permissive `channelPath` string handling
- workflow step accepting `requestBody`

These are intentional characterization tests for current behavior, not normative conformance examples.

## Notes

For each new negative fixture, I first verified an otherwise-valid baseline and then introduced the single invalidating change. This was done to ensure failures are attributable to the intended schema branch rather than unrelated document-shape issues.

## Validation

- Ran `npm test`
- Result: `62/62` tests passing

### DIFF
--- a/tests/schema/fail/async-depends-on-invalid-reference.arazzo.yaml
+++ b/tests/schema/fail/async-depends-on-invalid-reference.arazzo.yaml
@@ -1,0 +1,22 @@
+arazzo: 1.1.0
+info:
+  title: Async dependsOn invalid reference shape
+  version: 1.0.0
+sourceDescriptions:
+  - name: eventAPI
+    url: https://events.com/asyncapi.yaml
+    type: asyncapi
+workflows:
+  - workflowId: asyncWorkflow
+    steps:
+      - stepId: sendEvent
+        operationId: $sourceDescriptions.eventAPI.orderCreated
+        action: send
+      - stepId: receiveEvent
+        operationId: $sourceDescriptions.eventAPI.orderCreated
+        action: receive
+        dependsOn:
+          # Invalid on purpose: dependsOn accepts a plain local stepId (`sendEvent`),
+          # `$workflows.<workflowId>.steps.<stepId>`, or
+          # `$sourceDescriptions.<name>.<workflowId>.steps.<stepId>`, but not `$steps.<stepId>`.
+          - $steps.sendEvent
--- a/tests/schema/fail/workflow-missing-workflowId.arazzo.yaml
+++ b/tests/schema/fail/workflow-missing-workflowId.arazzo.yaml
@@ -1,0 +1,16 @@
+arazzo: 1.1.0
+info:
+  title: Workflow step must include workflowId
+  version: 1.0.0
+sourceDescriptions:
+  - name: userAPI
+    url: https://users.com/openapi.yaml
+    type: openapi
+workflows:
+  - workflowId: parentWorkflow
+    steps:
+      - stepId: nestedWorkflow
+  - workflowId: childWorkflow
+    steps:
+      - stepId: childStep
+        operationId: $sourceDescriptions.userAPI.listUsers
--- a/tests/schema/pass/characterization-workflow-step-with-requestBody.arazzo.yaml
+++ b/tests/schema/pass/characterization-workflow-step-with-requestBody.arazzo.yaml
@@ -1,0 +1,24 @@
+arazzo: 1.1.0
+info:
+  title: Characterization of workflow step with requestBody
+  version: 1.0.0
+sourceDescriptions:
+  - name: userAPI
+    url: https://users.com/openapi.yaml
+    type: openapi
+workflows:
+  - workflowId: parentWorkflow
+    steps:
+      # Characterization test: this passes today because requestBody is defined
+      # on step-object-base, so workflow steps currently accept it structurally.
+      # This was done to keep backward compatibility.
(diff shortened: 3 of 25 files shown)


frankkilcommins (approved):
Looks good, thanks @nashjain. Good to have more tests in this area.

It's good for all to remember the following statement in `~/Arazzo-Specification/src/schemas/validation/README.md`:

> As a reminder, the JSON Schema is not the source of truth for the Specification.
In cases of conflicts between the Specification itself and the JSON Schema, the
Specification wins. Also, some Specification constraints cannot be represented
with the JSON Schema so it's highly recommended to employ other methods to
ensure compliance.

tests/schema/fail/async-depends-on-invalid-reference.arazzo.yaml
```diff
@@ -0,0 +1,22 @@
...
+        operationId: $sourceDescriptions.eventAPI.orderCreated
+        action: send
+      - stepId: receiveEvent
+        operationId: $sourceDescriptions.eventAPI.orderCreated
+        action: receive
+        dependsOn:
+          # Invalid on purpose: dependsOn accepts a plain local stepId (`sendEvent`),
+          # `$workflows.<workflowId>.steps.<stepId>`, or
+          # `$sourceDescriptions.<name>.<workflowId>.steps.<stepId>`, but not `$steps.<stepId>`.
+          - $steps.sendEvent
```
DmitryAnansky, line 22: @frankkilcommins/ @nashjain 
I remember there was a discussion about what context $step.STEP_ID refers to.
Would $workflows.<workflowId>.steps.<stepId> be equivalent to $steps.<stepId> when referencing a step within the current workflow, or is there a difference in scope or resolution?

tests/schema/fail/workflow-missing-workflowId.arazzo.yaml
```diff
@@ -0,0 +1,16 @@
...
+  title: Workflow step must include workflowId
+  version: 1.0.0
+sourceDescriptions:
+  - name: userAPI
+    url: https://users.com/openapi.yaml
+    type: openapi
+workflows:
+  - workflowId: parentWorkflow
+    steps:
+      - stepId: nestedWorkflow
```
DmitryAnansky, line 12: Please check test name, it seems misleading, probably smth about empty step should be mentioned.

tests/schema/pass/characterization-workflow-step-with-requestBody.arazzo.yaml
```diff
@@ -0,0 +1,24 @@
...
+  title: Characterization of workflow step with requestBody
+  version: 1.0.0
+sourceDescriptions:
+  - name: userAPI
+    url: https://users.com/openapi.yaml
+    type: openapi
+workflows:
+  - workflowId: parentWorkflow
+    steps:
+      # Characterization test: this passes today because requestBody is defined
```
DmitryAnansky, line 12: Seems like this test is false positive and don't really check anything except the lack of possible verification. Do we need to have it still?

tests/schema/fail/async-depends-on-invalid-reference.arazzo.yaml
frankkilcommins: It would be the same, and the expectation is that evaluation would work similarly. When within the context of a workflow, `$steps.` refers to steps within the current workflow only. You could still use `$worklows.<workflowId>.steps` to get you to the same list assuming the <workflowId> is the current workflow.

tests/schema/fail/workflow-missing-workflowId.arazzo.yaml
nashjain: this test is actually about missing workflow id

tests/schema/pass/characterization-workflow-step-with-requestBody.arazzo.yaml
nashjain: toady workflow steps can contain requestBody, but ideally they should not.

PR #482, opened 2026-05-06
1.1.0 minor version

### LINKED ISSUE #253: Runtime expressions ABNF clarifications

### LINKED ISSUE #270: Feature: Add Support for AsyncAPI v2/v3 as Source Definition

### LINKED ISSUE #330: `goto` with `workflowId` for Success Actions and Failure Actions should support parameter mapping to Workflow Input Object
When configuring a Success Action Object or Failure Action Object of type `goto` which references a `workflowId`, then we should have a means to be able to supply the Input Object expected by that workflow.

Right not this is not support and/or the specification does not provide guidance on what to pass in to the referenced workflow.

### LINKED ISSUE #366: Proposal: Add support for JSONPath where JSON Pointer is supported

### LINKED ISSUE #423: Clarify truthy/falsy evaluation semantics for JSONPath conditions in Criterion Objects
> This topic was originally discussed in [slack](https://open-api.slack.com/archives/C022K8VD7AP/p1768999147946719)

The Arazzo specification allows for a `condition` to be expressed using JSONPath via the Criterion Object, but it does not currently define how the result of such expressions should be interpreted.

The general expectation is that conditions act as _truthy/falsy_ constructs, but without explicit guidance, tooling must infer whether a condition has passed or failed based on assumptions about the evaluation result.

A common example used in the spec, and existing documentation:

```
- context: $response.body
  condition: $[?count(@.pets) > 0]
  type: jsonpath
```

According to [RFC 9535](https://www.rfc-editor.org/rfc/rfc9535), the expression used above is valid and returns a **nodelist**, which may be empty or contain one or more nodes. It does not return a `boolean`.

The expected behaviour in Arazzo for the above condition:

- If the result is a non-empty nodelist, the condition _passes_
- If the result is an empty nodelist, the condition _fails_

However, this behaviour is currently not defined in the specification. There is:

- No normative language around evaluating the result of a JSONPath expression
- No mention of converting a NodesType to pass/fail using RFC 9535’s type conversion semantics (e.g. Section 2.4.2)

This omission leaves tooling authors guessing, potentially leading to inconsistent or non-compliant behavior across implementations.

In contrast, XPath-based conditions do allow for valid expressions that return boolean values directly (e.g., `count(/root/pets/*) > 0` could be used in the example for an equivalent XML response), making them more predictable for condition evaluation.

### LINKED ISSUE #424: Ambiguity in runtime expressions embedded in strings
The Arazzo Runtime Expression grammar's CHAR rule (specifically the `unescape` production) currently allows `{ (%x7B)` and `} (%x7D)` characters within expression names.

This is problematic because runtime expressions are embedded in template strings using `{expression}` syntax (e.g., `"client_id={$inputs.clientId}"`).

When `CHAR` permits these characters, an expression like `$inputs.foo}` becomes valid and consumes the closing brace,
making it impossible to reliably parse where an embedded expression ends. The fix is to exclude `{` and `}` from the `unescape` rule by changing `%x5D-10FFFF` to `%x5D-7A / %x7C / %x7E-10FFFF` (where %x7B is {, %x7C is |, and %x7D is }), ensuring the grammar correctly stops at expression boundaries.

`$request.body#/...` and `$response.body#/...` expressions with JSON pointers cannot be reliably extracted from `{expression}` syntax. This is because [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901) (JSON Pointer) allows the `}` character in pointer paths, making it impossible to determine where the expression ends.

### LINKED ISSUE #425: Introduction of ABNF for expressions embedded in strings
Hi,

I'm proposing to include and ABNF grammar to clearly determine how expressions embedded in strings are extracted.

```abnf
; Grammar for parsing template strings with embedded expressions
expression-string    = *( literal-char / embedded-expression )
embedded-expression  = "{" expression "}"
literal-char         = %x00-7A / %x7C / %x7E-10FFFF  ; anything except { (%x7B) and } (%x7D)
```

There is still an issue related to JSON Pointers documented in https://github.com/OAI/Arazzo-Specification/issues/424.

### LINKED ISSUE #426: Introduce complete grammar for parsing expressions

### LINKED ISSUE #427: Ambiguity in $sourceDescription.* runtime expressions

### LINKED ISSUE #428: Ambiguity in $components runtime expressions

### LINKED ISSUE #434: Feature: Add $self for identity-based referencing

### LINKED ISSUE #437: List or object usage in runtime expressions embedded in strings

### LINKED ISSUE #438: Proposal:  add support for `querystring` Parameter Format

### LINKED ISSUE #462: Update README in preparation for the 1.1 release
Things to include:

- include AsyncAPI in the opening paragraph
- update current version to `1.1.0` in the https://github.com/OAI/Arazzo-Specification#current-version---101 section
- update the image and add the soft version of the https://excalidraw.com/ meta to the repo.
- add tooling section before the "see in action section" and list current known tooling, but also point to openapi.tools for a more up-to-date community listing

### LINKED ISSUE #468: Add proposed release date to Arazzo 1.1-dev
Update the following table with a new line item for 1.1.0

## Appendix A: Revision History

| Version | Date | Notes |
| --- | --- | --- |
| 1.0.1 | 2025-01-16 | Patch release of the Arazzo Specification 1.0.1 |
| 1.0.0 | 2024-05-29 | First release of the Arazzo Specification |

frankkilcommins:
Arazzo - 1.1.0 release PR (to main)

Enhancements:
resolves #270
fixes #366
fixes #438
fixes #434

Bugs/Fixes:
fixes #253
fixes #330
fixes #424
fixes #425
fixes #426
fixes #428
fixes #437
fixes #427
fixes #423

House-keeping:
fixes #468
fixes #462


kevinduffey (approved):
LGTM

PR #499, opened 2026-07-06
Clarify when Criterion `context` is required

char0n:
## Problem

In the **Criterion Object** section, the `context` field description states:

> If `type` is specified, then the `context` MUST be provided (e.g. `$response.body` would set the context that a JSONPath query expression could be applied to).

Combined with the `type` field's rule:

> If omitted, then the condition is assumed to be `simple` […]

these create an unintended, contradictory edge case: if an author writes `type: simple` **explicitly**, the first clause demands that `context` MUST be provided — but a `simple` condition embeds its runtime expression directly in the `condition` string (e.g. `condition: $statusCode == 200`) and has no separate context to apply anything to. So the MUST-clause is meaningless (and impossible to satisfy usefully) for `simple`.

The spec's own **Simple Condition Example** confirms `simple` needs no context:

```yaml
- condition: $statusCode == 200
```

(no `context` field present).

The requirement is clearly *intended* to apply only to the condition types that apply a pattern/query to a context — `regex`, `jsonpath`, `xpath` — which is already stated separately in the `condition` field description:

> For regex or JSONPath, the `type` and `context` MUST be specified.

The four flavors differ exactly here: `simple` embeds runtime expressions inside the condition, whereas `regex`/`jsonpath`/`xpath` apply their pattern/query to a context defined by a runtime expression.

## Fix

Scope the `context` requirement to `regex`/`jsonpath`/`xpath`:

> If `type` is one of `regex`, `jsonpath`, or `xpath`, then `context` MUST be provided (e.g. `$response.body` sets the context that a JSONPath query expression could be applied to).

This removes the `type: simple` contradiction, aligns with the existing "For regex or JSONPath, the `type` and `context` MUST be specified" sentence, and preserves the original example.

This is a meaning-preserving editorial clarification (patch-eligible), targeting `v1.1-dev` as the earliest active branch where it is relevant.


ndenny (dismissed):

frankkilcommins (dismissed):
LGTM - reviewed in spec meeting

frankkilcommins:
@char0n could you review the JSON Schema too as I believe it's contradictory and sets context as dependentRequired if type is specified (and not only when the value of type is one of the three that we expect context to be provided).

```
  criterion-object:
    $comment: https://spec.openapis.org/arazzo/v1.1#criterion-object
    description: |-
      An object used to specify the context, conditions, and condition types
      that can be used to prove or satisfy assertions specified in Step Object successCriteria,
      Success Action Object criteria, and Failure Action Object criteria
    type: object
    properties:
      context:
        description: A runtime expression used to set the context for the condition to be applied on
        type: string
      condition:
        description: The condition to apply
        type: string
      type:
        description: The type of condition to be applied or a reference to an expression type object
        oneOf:
          - type: string
            enum:
              - simple
              - regex
              - jsonpath
              - xpath
            default: simple
          - $ref: '#/$defs/expression-type-object'
    required:
      - condition
    dependentRequired:
      type:
        - context
```

frankkilcommins:
@char0n gentle ping to see if you'd like to review the possible JSON Schema contradiction

char0n:
@frankkilcommins done, sorry for delay

PR #516, opened 2026-07-09
fix(spec): specify ECMA-262 dialect for regex Criterion condition type

### LINKED ISSUE #515: Criterion Object: regex condition type does not specify a regular expression dialect
### Summary
The Criterion Object's `regex` condition type does not specify which regular-expression dialect implementations must use, while the sibling `jsonpath` and `xpath` types both pin a standard. This leaves regex matching non-portable across Arazzo runners.

### Details
In the Criterion Object (Arazzo 1.0.1 and 1.1.0), the `type` field allows `simple`, `regex`, `jsonpath`, or `xpath`. For two of these, the spec pins a standard:
- `jsonpath` → RFC 9535 lineage (`draft-goessner-dispatch-jsonpath-00`), versioned via the Criterion Expression Type Object
- `xpath` → XML Path Language 3.1, also versioned

But for `regex`, the spec says only: *"regex - where a regex pattern is applied on the supplied context."* No dialect is named (ECMA-262, I-Regexp/RFC 9485, PCRE, XSD, RE2, …), and `regex` has no entry in the Criterion Expression Type Object, so there is no way to declare which flavor is intended.

### Why this matters
Regex dialects differ in ways that change matching results — lookbehind `(?<=…)`, named groups, backreferences, Unicode property escapes `\p{…}`, and even the meaning of `.`. A condition such as `condition: '(?<=x)y'` or `'\p{L}+'` may pass on one runtime and fail (or error) on another. This is the same class of portability gap the JSONPath working group explicitly set out to solve.

Notably, the two specifications closest to Arazzo made **different** explicit choices:
- **OpenAPI** (Arazzo's sibling under the OAI) pins **ECMA-262** for `pattern`/`patternProperties`, inherited from JSON Schema Validation Draft 2020-12 §6.3.3 ("SHOULD be a valid regular expression, according to the ECMA-262 regular expression dialect"). OpenAPI also authors its own normative field constraints (e.g. Components key pattern `^[a-zA-Z0-9\.\-_]+$`) in ECMA-262 syntax.
- **JSONPath** (RFC 9535 — which Arazzo already references for its `jsonpath` type) pins **I-Regexp** (RFC 9485), an interoperability-focused subset of XSD regex, for its `match()`/`search()` functions.

So Arazzo currently references a spec (RFC 9535) that uses I-Regexp, while its sibling (OpenAPI) uses ECMA-262 — and Arazzo's own `regex` type commits to neither.

### Proposed resolution
Specify a dialect for the `regex` type. Preferred: **ECMA-262**, matching OpenAPI/JSON Schema (the sibling specification, and what the most common runners implement via native regex engines). A single clarifying sentence in the `regex` description would suffice, e.g.: *"The condition MUST be a valid regular expression according to the ECMA-262 regular expression dialect."*

Alternatively, add a `regex` entry to the Criterion Expression Type Object with a `version` field (mirroring `jsonpath`/`xpath`) if multiple dialects should be selectable — though for regex a single pinned dialect is likely simpler and sufficient.

### Affected versions
Arazzo 1.0.0, 1.0.1, and 1.1.0.

char0n:
Closes #515

## Summary

The Criterion Object's `regex` condition type did not name a regular-expression dialect, while the sibling `jsonpath` (RFC 9535) and `xpath` (XML Path Language 3.1) types both pin a standard. This left regex matching non-portable across Arazzo runners: dialects differ in lookbehind `(?<=…)`, named groups, backreferences, and Unicode property escapes `\p{…}`, so a condition such as `(?<=x)y` or `\p{L}+` could pass on one runtime and fail (or error) on another.

This PR pins **ECMA-262** for the `regex` type, matching OpenAPI / JSON Schema Validation (the sibling specification under the OAI, which pins ECMA-262 for `pattern`/`patternProperties`) and what the most common runners implement via native regex engines.

## Changes

Two clarifying edits to `src/arazzo.md`, both linking to the [ECMA-262 patterns grammar](https://tc39.es/ecma262/#sec-patterns):

- Criterion Object condition flavors list — the `regex` entry now states the pattern MUST be a valid regular expression according to the ECMA-262 dialect.
- Regex Conditions evaluation section — the `condition` MUST be a valid ECMA-262 regular expression.

This is a clarification that does not change the intended meaning of the specification, so it targets `v1.1-dev` per the [branching strategy](https://github.com/OAI/Arazzo-Specification/blob/main/CONTRIBUTING.md#branching-strategy) (apply to the earliest relevant active branch). `markdownlint` passes.

### DIFF
--- a/src/arazzo.md
+++ b/src/arazzo.md
@@ -838,7 +838,7 @@ An object used to specify the context, conditions, and condition types that can 
 There are four flavors of conditions supported:
 
 - simple - where basic literals, operators, and loose comparisons are used in combination with [Runtime Expressions](#runtime-expressions).
-- regex - where a regex pattern is applied on the supplied context. The context is defined by a [Runtime Expression](#runtime-expressions).
+- regex - where a regex pattern is applied on the supplied context. The pattern MUST be a valid regular expression according to the [ECMA-262](https://tc39.es/ecma262/#sec-patterns) regular expression dialect. The context is defined by a [Runtime Expression](#runtime-expressions).
 - jsonpath - where a JSONPath expression is applied. The root node context is defined by a [Runtime Expression](#runtime-expressions).
 - xpath - where an XPath expression is applied. The root node context is defined by a [Runtime Expression](#runtime-expressions).
 
@@ -954,7 +954,7 @@ Example:
 
 ###### Regex Conditions
 
-When `type` is `regex`, the `condition` MUST be a valid regular expression pattern, and `context` MUST be provided. The condition evaluates to:
+When `type` is `regex`, the `condition` MUST be a valid regular expression according to the [ECMA-262](https://tc39.es/ecma262/#sec-patterns) regular expression dialect, and `context` MUST be provided. The condition evaluates to:
 
 - A condition passes (truthy) when the regex pattern matches the `context` value.
 - A condition fails (falsy) when the regex pattern does not match the `context` value.


handrews:
Please note that as of this week, in the JSON Schema IETF working group we are considering settling on I-REGEXP as a more feasible common ground.  ECMA262 support, while technically recommended (SHOULD) by JSON Schema, is not, in pracitce, widely correctly supported.

frankkilcommins:
@char0n as OpenAPI are considering I-REGEXP RFC 9485 as a more feasible common ground than ECMA262, it might be sensible to follow suit here with recommendation/clarification.

We'll discuss in Arazzo call today

PR #533, opened 2026-07-27
feat(spec): add SOAP support

frankkilcommins:
Adds first-class WSDL support to Arazzo 1.2 via a minimal, additive set of changes. Because SOAP operates over HTTP, the vast majority of Arazzo already works for SOAP workflows without modification. Only two things were broken: no valid source description type for WSDL, and no valid way to reference a WSDL operation from a step.

**Changes**

- Source Description Object: adds "wsdl" to the type enum (WSDL 1.1 and WSDL 2.0 supported); updates introductory prose and adds a WSDL example
- Step Object: extends operationId to match WSDL operation names (wsdl:operation/@name, required in both WSDL versions); adds a MUST NOT use operationPath for WSDL steps
- New subsection - Authoring Steps for WSDL Source Descriptions: covers HTTP method/endpoint resolution from the WSDL binding, SOAPAction/Content-Type guidance for SOAP 1.1 and 1.2, SOAP Fault detection patterns, and request body envelope requirements
- Runtime Expressions: extends source description resolution to cover WSDL operation name matching (WSDL 1.1 portType, WSDL 2.0 interface); adds WSDL example; updates ABNF comments
- Schema: adds "wsdl" to source type enum; renames openapi-step-object → openapi-or-wsdl-step-object
- Examples: adds soap-customer-crud.arazzo.yaml (WSDL-only) and hybrid-order-fulfillment.arazzo.yaml (OpenAPI + WSDL chained workflow)
- Tests: adds pass/fail/characterization schema tests for WSDL-specific cases

**Design note:** operationId reuse vs operationName

WSDL steps use operationId to reference operations by name rather than introducing a new operationName field. The trade-off: because operationId is structurally identical between  OpenAPI and WSDL steps, the schema cannot enforce the operationPath prohibition for WSDL steps at validation time. This is instead enforced via spec prose (MUST NOT) and is a candidate for a linting rule. Keeping operationId avoids inflating the step object with a field that serves the same semantic purpose, consistent with how AsyncAPI steps also reuse operationId alongside their own unique fields.


kevinduffey (dismissed):
LGTM. dicussion on slack, otherwise fine.

examples/1.2.0/hybrid-order-fulfillment.arazzo.yaml
```diff
@@ -0,0 +1,136 @@
...
+            </soap:Envelope>
+        successCriteria:
+          - condition: $statusCode == 200
+          - context: $response.body
+            condition: //CheckInventoryResponse/Available[text()='true']
+            type:
+              type: xpath
+              version: xpath-31
+        outputs:
+          availableStock: //CheckInventoryResponse/AvailableQuantity
```
DmitryAnansky: Is it a valid output selector?
Should there be some additional information about type?
e.g.
```
        context: $steps.fetchXml.outputs.invoiceXml
        selector: /Invoice/Header/InvoiceNumber
        type:
          type: xpath
          version: xpath-30
```

frankkilcommins: Thanks @DmitryAnansky - fixed and committed.

examples/1.2.0/hybrid-order-fulfillment.arazzo.yaml
```diff
@@ -0,0 +1,145 @@
...
+    API handles product lookup and order creation, while a legacy SOAP/WSDL
+    fulfillment service checks inventory availability and dispatches the order.
+    Outputs from REST steps are chained into SOAP steps and vice versa.
+
+sourceDescriptions:
+  - name: retailAPI
+    url: https://api.example.com/retail/openapi.yaml
+    type: openapi
+  - name: fulfillmentService
+    url: https://legacy.example.com/fulfillment?wsdl
```
DmitryAnansky: This link doesn’t point to an example. Would it make sense to include this description in the repository or update the link to point to a valid example instead?

examples/1.2.0/soap-customer-crud.arazzo.yaml
```diff
@@ -0,0 +1,195 @@
...
+sourceDescriptions:
+  - name: customerService
+    url: http://localhost:4010/customer-service?wsdl
+    type: wsdl
+
+workflows:
+  - workflowId: soap-crud
+    steps:
+      - stepId: list-customers
+        operationId: $sourceDescriptions.customerService.ListCustomers   # FIXED: was operationPath: POST /api/...
```
DmitryAnansky: Do you need comment here?

src/schemas/validation/schema.yaml
DmitryAnansky, line 266: Does wsdl have operationPath?
Maybe its better to have separate type for wsdl-step-object ?

examples/1.2.0/hybrid-order-fulfillment.arazzo.yaml
frankkilcommins: Fair point - i can add local files to make it easier to understand, follow, and validate

src/schemas/validation/schema.yaml
frankkilcommins: I can create a wsdl-step-object for clarity purposes. However as there's no unique distinction from a schema validation perspective between an openapi step and a wsdl step (e.g. the both will use operationId), I'll have to change the polymorphic construct on the step definition to an `anyOf`. We won't be able to enforce not using an `operationPath` for wsdl use cases but at least the schema will be clear in its intent for readers (human or agent).

tests/schema/pass/characterization-wsdl-step-with-operationPath.arazzo.yaml
```diff
@@ -0,0 +1,17 @@
...
+    type: wsdl
+workflows:
+  - workflowId: basicWorkflow
+    steps:
+      # Characterization test: this passes today because the schema does not
+      # resolve sourceDescriptions.type when selecting the step-object branch.
+      # operationPath MUST NOT be used with a wsdl source description per spec
+      # prose, but the schema cannot enforce this cross-object constraint.
+      - stepId: listCustomers
+        operationPath: $sourceDescriptions.customerService#/paths/~1ListCustomers
```
DmitryAnansky, line 25: Could you please clarify why there should be a false-positive test case that passes even though WSDL does not define operationPath?

frankkilcommins: This is an intentional characterisation test that documents a known gap in the JSON Schema validation - the schema cannot enforce the operationPath _MUST NOT_  constraint for wsdl source descriptions because doing so would require resolving the sourceDescription type at validation time. 

The specification prose remains the authoritative source of truth, and conformance tooling is responsible for enforcing this rule at runtime. The test exists so that this limitation is explicitly acknowledged rather than silently undocumented (and to perhaps prevent folks raising issues against the schema for it).

I'll add more explicit documentation within the test file.

kevinduffey (dismissed):
LGTM

ndenny (dismissed):
This looks great, straightforward and fits into Arazzo neatly! 

I don't envy anyone trying to work out XPaths for SOAP responses, but hopefully tooling will save us there 🤣

nashjain:
Sorry for joining the party pretty late.

I'm not thrilled about changing `oneOf` to `anyOf`. Technically it should be `oneOf`, we cannot have more than 1 at a time. I do understand why you had to change it to `anyOf`.

One potential solution to address this issue would be replace `operationId` with `operationName` for WSDL. As you pointed out, WSDLs do not have `operationId`, they have `operationName`.  `operationName`  can become a clear differentiator for the wsdl-step-object. So then we can go back to `oneOf`.

frankkilcommins:
@nashjain 

> Sorry for joining the party pretty late.

You're always welcome!

> I'm not thrilled about changing `oneOf` to `anyOf`. 

Nor am I, but it was weight that up with the addition of another step fixed field

> One potential solution to address this issue would be replace `operationId` with `operationName` for WSDL. 

That is actually what I was proposing first but then changed back to reuse existing fixed field (`operationId`). I am fine to change to `operationName` if that's the broader wish (cc @kevinduffey, @ndenny). In one way it's very clear as it aligns with WSDL's own terminology.

A longer term broader piece of work might be to revisit what we toyed with as part of the AsyncAPI PR and introduce a step `type` discriminator.

nashjain:
> > One potential solution to address this issue would be replace `operationId` with `operationName` for WSDL.
> 
> That is actually what I was proposing first but then changed back to reuse existing fixed field (`operationId`). I am fine to change to `operationName` if that's the broader wish (cc @kevinduffey, @ndenny). In one way it's very clear as it aligns with WSDL's own terminology.

Agree. You can also see #556 in which we've used a grpcMethod so the step body is unique.
> 
> A longer term broader piece of work might be to revisit what we toyed with as part of the AsyncAPI PR and introduce a step `type` discriminator.

Agreed. This was what we had first proposed at part of #367. stepType discriminator would keep it clean and very declarative. But now if we have to add it, it would be a backward breaking change.

frankkilcommins:
@ndenny @kevinduffey @nashjain - I've introduced `operationName` and reverted the schema to a `oneOf` construct for step object

nashjain:
> @ndenny @kevinduffey @nashjain - I've introduced `operationName` and reverted the schema to a `oneOf` construct for step object

LGTM

ndenny (approved):

kevinduffey (approved):
LGTM

PR #556, opened 2026-08-28
feat(spec): add Protocol Buffer RPC support

nashjain:
Adds first-class RPC support to Arazzo 1.2 using Protocol Buffer service descriptions. The proposal introduces a protocol-neutral RPC Step model with protocol-specific execution semantics for gRPC, gRPC-Web, Twirp, and Connect.

Protocol support

- gRPC: unary, server-streaming, client-streaming, and bidirectional-streaming RPCs.
- gRPC-Web: unary and server-streaming RPCs.
- Twirp: unary RPCs.
- Connect: unary, server-streaming, client-streaming, and bidirectional-streaming RPCs.

Changes

- Source Description Object: adds `"protobuf"` as a source type for Protocol Buffer source documents.
- Step Object: adds `rpcMethod` for referencing methods using `<fully-qualified-service-name>/<method-name>`.
- Protocol discriminator: adds the required `rpcProtocol` field with `"grpc"`, `"grpc-web"`, `"twirp"`, and `"connect"` as supported values.
- Method validation: defines the lexical structure of source-qualified RPC method references. Source-aware validation resolves the referenced service and method from the Protocol Buffer source.
- Message representation: uses ProtoJSON for Protocol Buffer values exposed in Arazzo documents and runtime expressions.
- Request bodies: defines protocol-specific request-body shapes, content types, streaming cardinality, message ordering, and empty-stream behavior.
- Protocol semantics: documents routing, wire encoding, completion, cancellation, timeouts, status, errors, metadata, and trailing metadata for each supported protocol.
- Parameters: adds `"metadata"` as a parameter location for protocols that support RPC metadata.
- Runtime Expressions: adds request metadata, response metadata, trailing metadata, and protocol-specific status expressions such as `$response.status#/code`.
- Success criteria: makes protocol outcomes available through runtime expressions without introducing implicit RPC success criteria; examples define their expected success explicitly.
- Examples: adds standalone examples for each RPC protocol and a hybrid OpenAPI and RPC workflow.
- Schema: adds protocol-discriminated RPC Step definitions and protocol-specific request-body validation.
- Tests: adds passing and failing schema coverage for protocol discrimination, RPC method references, request-body forms, content types, streaming restrictions, metadata, and mutually exclusive Step fields.

Design notes

`rpcMethod` and `rpcProtocol` are intentionally separate. The method identifies a declaration in a Protocol Buffer source, while the protocol determines how that method is invoked and how protocol-specific status, errors, metadata, streaming, and wire behavior are interpreted.

Client-streaming and bidirectional-streaming request messages are represented as authored ordered arrays. Interactive bidirectional sequencing—where a later outbound message depends on an earlier inbound message—is outside the scope of this proposal.


src/arazzo.md
```diff
@@ -414,31 +434,131 @@ outputs:
...
+| <a name="stepGrpcMethod"></a>grpcMethod           |                                         `string`                                         | A fully-qualified gRPC method identifier from a Protocol Buffer service declaration, in the form `$sourceDescriptions.<name>.<fully-qualified-service-name>/<method-name>`; the logical identifier is `<fully-qualified-service-name>/<method-name>` (e.g., `org.v1.Library/GetBook`). This field is mutually exclusive with `operationId`, `operationPath`, `channelPath`, and `workflowId` and only applicable for `grpc` steps.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
```
frankkilcommins: Should we consider `rpcMethod` rather than `grpcMethod`? The `protobuf` source type already signals the technology; locking the  field name to gRPC prevents reuse for other RPC frameworks.

Additionally the `/` separated service name and method name could be assumed to be JSON Pointer notation. I guess your leveraging the wire-level HTTP/2 path convention, but I am wondering if it's more explicit if we use the protobuf fully qualified name in dot notation? E.g. _package.ServiceName.MethodName_ which would be something like `$sourceDescriptions.libraryService.google.pubsub.v2.PublisherService.CreateTopic`.

Do we need any notes or recommendations if a service is defined in an imported .proto? Perhaps we _SHOULD_ require a dedicated source description for that file rather than referencing it transitively?

src/arazzo.md
```diff
@@ -414,31 +434,131 @@ outputs:
...
+```
+
+is visible in Arazzo as:
+
+```yaml
+bookId: ...
+chunksReceived: ...
+```
+
+unless an explicit protobuf `json_name` overrides the field name. Runtime expressions therefore use `$response.body#/bookId` and `$response.body#/chunksReceived`. See the [ProtoJSON specification](https://protobuf.dev/programming-guides/json/) for detailed conversion semantics.
```
frankkilcommins: Perhaps word as follows rather than the sentence fragment:

> The default JSON field name is the lowerCamelCase version of the proto field name. An explicit `json_name` field option overrides this default.

src/arazzo.md
```diff
@@ -1355,13 +1519,24 @@ The runtime expression is defined by the following [ABNF](https://tools.ietf.org
       "$self"
   )
 
-  ; Request/Response sources
-  source = ( header-reference / query-reference / path-reference / body-reference / payload-reference )
+  ; Request sources
+  request-source = ( header-reference / query-reference / path-reference / body-reference / payload-reference / request-metadata-reference )
+
+  ; Response sources
+  response-source = ( header-reference / body-reference / payload-reference / status-reference / response-metadata-reference )
```
frankkilcommins, line 1676: possibly should flag that using `status-reference` and `response-metadata-reference` are only meaningful for gRPC steps. Using on OpenAPI for example is undefined behaviour

frankkilcommins (commented):
Great work @nashjain.

I have a few comments, nothing ground breaking on first review.

src/arazzo.md
nashjain: Thanks, @frankkilcommins . I’m in the process of addressing the first point by generalizing this to `rpcMethod`, with `rpcProtocol` acting as the discriminator for the different RPC protocols that use Protobuf service definitions.

For method identity, I’m inclined to retain the `$sourceDescriptions.<source>.<fully-qualified-service>/<method>` format. Dot notation is the canonical Protobuf descriptor full name, whereas the slash form is the familiar RPC method notation used by gRPC and the other protocols under consideration. It also makes the boundary between the fully qualified service name and the method name explicit. We can clarify that this is an RPC method identifier, not a JSON Pointer; a JSON Pointer is either empty or begins with /, while this value begins with a source-description reference.

For imported definitions, I believe a Protobuf Source Description should act as the entry point to the complete Protobuf dependency graph. Methods declared in directly or transitively imported files should therefore be referenceable without requiring separate Source Descriptions for those files. Requiring every imported file to be declared separately would expose the physical Protobuf file structure in the Arazzo document. It would also require unnecessary Arazzo changes when declarations move between files without changing their fully qualified identities.

src/arazzo.md
nashjain: Fixed

src/arazzo.md
nashjain: Fixed

src/arazzo.md
nashjain: `grpcMethod` is now changed to `rpcMethod`.

PR #567, opened 2026-09-02
feat(spec): add GraphQL operation support

nashjain:
## Summary

Adds GraphQL operation support to the Arazzo 1.2 specification.

This enables workflows to reference GraphQL schemas and executable documents, execute inline or externally defined operations, supply GraphQL variables and request extensions, and expose GraphQL execution results through existing Arazzo runtime expressions.

## Changes

- Add `graphql` as a Source Description type.
- Add the GraphQL Operation Object and `graphqlOperation` Step field.
- Support inline GraphQL operations and references to named operations in external Executable Documents.
- Define GraphQL variable values using Parameter Objects with `in: variable`.
- Add `valueMode` to distinguish literal object values from Selector Objects.
- Add separate `extensions` and `extensionsSelector` fields to remove ambiguity between literal extension maps and selectors.
- Define operation selection, GraphQL validation, variable coercion, request-error, and execution-error behavior.
- Define support for finite subscriptions that complete naturally within the required Step timeout.
- Expose query, mutation, and subscription results through `$response.body`.
- Extend Runtime Expressions to reference complete GraphQL source descriptions and named operations.
- Add GraphQL examples, schema validation fixtures, SDL, and reusable operation documents.

## Scope boundaries

- GraphQL Steps identify the schema and executable operation, but not a concrete deployment.
- Endpoint, transport protocol, authentication, schema discovery, and client construction are supplied through implementation-defined execution-environment configuration.
- Unbounded subscriptions, receive-until conditions, and successful early unsubscribe are outside the supported execution model.

## Validation

- All 78 schema tests pass.
- Markdown validation passes.
- `git diff --check` passes.

### DIFF
--- a/examples/1.2.0/graphql.arazzo.yaml
+++ b/examples/1.2.0/graphql.arazzo.yaml
@@ -1,0 +1,97 @@
+arazzo: 1.2.0
+info:
+  title: GraphQL book workflows
+  version: 1.0.0
+sourceDescriptions:
+  - name: shopSchema
+    type: graphql
+    url: ./schema.graphqls
+  - name: shopOperations
+    type: graphql
+    url: ./operations.graphql
+workflows:
+  - workflowId: lookupBook
+    inputs:
+      type: object
+      properties:
+        bookId:
+          type: string
+        requestId:
+          type: string
+    steps:
+      - stepId: getBook
+        graphqlOperation:
+          schema: $sourceDescriptions.shopSchema
+          operation: |
+            query GetBook($id: ID!) {
+              book(id: $id) {
+                id
+                title
+              }
+            }
+          extensions:
+            tracing:
+              requestId: $inputs.requestId
+        parameters:
+          - name: id
+            in: variable
+            value: $inputs.bookId
+        outputs:
+          title: $response.body#/data/book/title
+  - workflowId: updateBook
+    inputs:
+      type: object
+      properties:
+        bookId:
+          type: string
+        city:
+          type: string
+        country:
+          type: string
+        operationHash:
+          type: string
+    steps:
+      - stepId: updateBook
+        graphqlOperation:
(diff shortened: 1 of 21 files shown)


examples/1.2.0/graphql.arazzo.yaml
```diff
@@ -0,0 +1,97 @@
+arazzo: 1.2.0
```
wtrocki, line 1: Discussed offline: Adding example of extension to the graphql query.

nashjain: I'm sorry, I mixed up the intent of extension when we discussed on the call.

The GraphQL Specification reserves the extensions map for implementation-specific additional information, allowing servers and clients to extend the protocol.  This mechanism is commonly used for tracing and performance monitoring.

Before we jump into the solution, I would like to step back and understand the problem statement fully. 

From what I remember, on the call you mentioned that you would like the same GraphQL query in an Arazzo workflow graphql step to return different fields and you were looking for a way to do that without changing the workflow. Am I right?

In my view, in a given Arazzo workflow for a specific graphql step, you would always require the same set of fields. However in a different step using the same query you might want different fields. Is this understanding correct?

wtrocki: > In my view, in a given Arazzo workflow for a specific graphql step, you would always require the same set of fields. However in a different step using the same query you might want different fields. Is this understanding correct?

Correct -- thank you for the explanation.

For clarity: each step should explicitly define its GraphQL query, so the agentic workflow is guardrailed to a specific response set.

nashjain: > For clarity: each step should explicitly define its GraphQL query, so the agentic workflow is guardrailed to a specific response set.

Yes, that is correct. I've designed it keeping the same in mind.

PR #568, opened 2026-09-03
feat(spec): add actor-in-the-loop support

frankkilcommins:
Adds _actor-in-the-loop_ interaction steps to Arazzo 1.2, enabling workflows to suspend execution at a defined pause point until a human, AI agent, or external automated system provides input before proceeding.

**Changes:**

**Step Object:** 
 - adds `interaction` field (mutually exclusive with operationId, operationPath, channelPath, and  workflowId)  
 - adds `onTimeout` - an ordered list of failure actions evaluated when timeout expires, applicable to all step types  
 - adds `onCancel` evaluated when an actor explicitly abandons an interaction step 
 - extends timeout to accept ISO 8601 duration strings (e.g. "PT8H", "P2D") in addition to integer milliseconds.

**Interaction Object** (New): 
Describes the pause point itself with five fixed fields (prompt, context, inputSchema, mode, and the redirect-specific triad operationId / operationPath / url). Defines three interaction modes `form`, `redirect`, `acknowledge`

**Components Object:** 
Adds interactions map for reusable Interaction Objects, referenced via `$components.interactions.<name>.`

**Runtime Expressions:** 
Adds `$interaction.payload`, `$interaction.payload#/<json-pointer>`, and `$interaction.query.<name>`. Updates ABNF with interaction-source production and adds "interactions" to component-type; adds example rows to the expressions table.

**Schema:** 
Adds `interaction-step-object` and `interaction-object definitions`. Extends `step-object` oneOf. Updates `step-object-base` (`timeout` broadened, `onTimeout` added). Updates components-object with interactions map. All `$comment URLs` updated to v1.2. `inputSchema` is enforced as required for form and redirect modes via an allOf / if/then construct. Acknowledge mode is exempt as the spec defines its default schema.

Added rich examples and tests to cover the extensions. All existing tests still pass.

closes #449


SensibleWood:
Just registering my approval for the JAR/PAR examples in the PR 👍 good one

src/arazzo.md
```diff
@@ -1331,6 +1336,173 @@ An XPath example using older XPATH 3.0:
...
+      type: object
+      required: [approved]
+      properties:
+        approved: { type: boolean }
+        notes:    { type: string }
+  timeout: PT8H
+  onTimeout:
+    type: goto
+    stepId: cancel-deployment        
+  successCriteria:
```
DmitryAnansky, line 1442: Should onSuccess be mandatory here?
It looks like, after this step is executed successfully, the next step in the queue can be triggered.
That next step could be something like the described `cancel-deployment`, which might break the intended execution flow and logic.

frankkilcommins: Hey @DmitryAnansky,

I agree that this example is incomplete. I will fix it up and align it more with ones I have in the examples folder.

Thanks for highlighting

src/arazzo.md
```diff
@@ -1331,6 +1336,178 @@ An XPath example using older XPATH 3.0:
...
+| <a name="interactionInputSchema"></a>inputSchema    |                                    JSON Schema Object                                  | A [JSON Schema 2020-12](https://json-schema.org/specification) object describing the structure and constraints of the expected actor response. The field MUST be specified for `form` and `redirect` modes. Executor implementations SHOULD validate actor input against this schema before resuming the workflow. For `redirect` mode, this schema describes the expected callback query parameters. For `acknowledge` mode, when omitted the default schema `{ "acknowledged": boolean }` applies.                             |
```
ndenny, line 1351: A semantic/taste question - I feel like this is the `outputSchema` as it's what the interaction outputs, not what it takes as inputs?

frankkilcommins: @ndenny thanks. 

Regarding `inputSchema` naming, the thought process is that it describes what the step must _receive_ as input from the actor (and from the perspective of the workflow), which is consistent with how inputs works at workflow level. 

But it's good to have some discussion on the naming.

kevinduffey:
Frank, I've been experimenting with actor-in-the-loop for a while and wrote up my own draft for 1.2 that I never posted. Yours covers things mine didn't (`onCancel`, `prompt` and `context`, and redirect mode). Here's what I think mine can add to this one.

### State across pause and resume

The Resumption section leaves storage, and what goes into `workflowState`, to the executor, and I agree with that. I'd still define "sufficient state" from the outside: after a resume, everything the rest of the workflow can reference (`$inputs`, outputs of completed steps, retry counts, where it was) evaluates the same as when it paused. Whether that state lives inside `workflowState` or in the executor's own storage is up to the tool.

Two SHOULDs I'd add:

- When the suspension is stored durably, which your text already recommends for long-lived interactions, store it before the interaction is presented or the redirect goes out, so nobody acts on an interaction a crash already lost.
- Refuse to resume if the workflow definition changed while it was waiting, rather than continuing against steps that aren't the ones it paused in. Recording a hash of the workflow at suspension and checking it on resume is enough.

A pause shouldn't be treated as the execution ending. Whatever comes after the step waits for it, the same as for any other step, and the workflow isn't finished while an interaction is open.

### Async work still in flight during a pause

This one sits between the 1.1 async text and this PR, and neither covers it. Say a workflow sends an order, has a `receive` with a `correlationId` waiting on the order-created message, then pauses for an approval, and a step after the approval `dependsOn` that `receive`. The message can easily arrive while the workflow is paused. I'd pin down:

- The message is kept and matched when the workflow resumes, not dropped because nothing was running when it came in. What the executor needs to match it (the correlation and the deadline) is part of the suspended state.
- The `receive` timeout keeps running in wall-clock time during the pause, and a message counts if it arrived before the deadline, even if the workflow only resumes after it.
- A message arriving doesn't resume the workflow. Only resolving the interaction does.

One I don't have a firm answer for: if the `receive` times out during the pause and its `onFailure` ends the workflow, does that happen right away (closing the open interaction), or only when the workflow reaches the step that depends on it? Right away stops someone from approving an order that already failed. Waiting keeps the document's step order. Two executors could reasonably pick either, so it's worth saying which. Should we consider a "pause" in any async in flight step timeout period and resume those (at the tool level) when the workflow resumes so that the timeout "continues" after resume?

### One token, one interaction

Keeping `workflowState` as the one standard name, opaque and executor-defined, makes sense to me, and your text already has it generated at suspension time. I'd spell out what that means when a step runs more than once. A step can be reached again through `retry` or `goto`, and a late answer from the first pass must not land on the second. So: each `workflowState` identifies one interaction and is never reused; an answer to one that already finished, timed out, or was cancelled is rejected, even if the same step has a newer one open; and a token can't start an execution or resume a different one.

This is easy to get wrong. Matching the resume on the step catches an answer meant for a different step, but not one meant for an earlier pass through the same step. A fresh random handle per interaction avoids it.

Requiring HMAC or AEAD protection is right for forgery and tampering. It proves the executor issued the token, though, not that whoever holds it is allowed to answer. I'd add that holding `workflowState` MUST NOT by itself authorize a response, unless a deployment deliberately treats it as a bearer credential over a confidential channel.

### Timeouts and bad answers

With timeouts in hours or days (your change-approval example waits 8 hours), I'd pin down four things: the deadline starts at suspension, it survives a restart when the suspension is stored durably, answers after it are refused, and an answer arriving right at the deadline resolves atomically so `onTimeout` and the success path never both run.

For form mode I'd validate at submission instead of leaving it implementation-defined. An invalid response goes back to the sender with the errors, the interaction stays open, and the workflow never sees it. Then `onFailure` means the answer was no, not that someone mistyped a field.

### Who answers: actor and interaction

My draft had the step point at an `actorId` instead of carrying an `interaction`. I don't think it has to be one or the other. Your Interaction Object describes the moment: what to show, how to collect it, what comes back. Nothing describes who is answering, and that's what routing and authorization key off. In `change-approval`, `secondaryApprover` is a required input that nothing uses, and the escalation prompt says "escalated to you" with no standard way to say who "you" is.

So I'd add an actor, defined once and referenced from the interaction by `actorId`. It has a `role` naming who is expected to act (a group or queue, never a URL or address), which the executor maps to real people through its own config, and an `outputSchema` for what they send back. +1 to @ndenny on that name, since it's what comes out of the interaction, not what goes into it. `interaction` stays on the step for the prompt, context, mode, and redirect target.

```yaml
actors:
  - actorId: changeApprover
    role: change-approvers
    outputSchema:
      type: object
      required: [approved]
      properties:
        approved: {type: boolean}

# in a workflow's steps
- stepId: request-primary-approval
  interaction:
    actorId: changeApprover
    prompt: "Change {$inputs.changeId} is ready for {$inputs.environment}. Approve?"
  timeout: PT8H
```

This also answers the reuse question from our meeting today (9/16/2026). You're right that the prompt and context belong to the step: they point at that workflow's inputs and step outputs, and a Reusable Object can't override them, which is why the `interaction-in-components` test fixture ends up with a generic "Please approve this request." But who answers and what they send back do repeat. In `change-approval` both approval steps have the identical response schema with different prompts. Splitting it lets that part be reused and keeps the contextual part on the step. An interaction without `actorId` works the way it does now, so none of your examples change.

